PR #3, opened 2023-03-14
DPR-391 introduce micronaut and make use of dependency injection

carwynmoj:
Summary of changes
* introduced micronaut to manage dependency injection
* annotate classes as appropriate in order to make use of dependency injection
* revised the `KinesisTestJob` to make use of the micronaut application context and demonstrate how to set up a job that uses the `PicolocliRunner`
* added tests for the `Properties` class
* configured the jacoco test coverage plugin which will run by default when the test task is run
* removed the placeholder task and test since this is no-longer required
* other minor changes including updating the readme


rambabu-posa (approved):
LGTM

dominic-messenger-moj (commented):
Looks good.

PR #4, opened 2023-03-15
DPR-319 add logging dependencies and configuration for local dev

carwynmoj:
Summary of changes
* initially I'd set up logback but discovered that when run as a glue job we actually use the log4j2 binder so flipped the libraries and configuration over to that
* replaced all printlns with logging at info
* provided a config for local development only - glue provides its own log4j2 properties file which overrides anything we bundle in the jar


dominic-messenger-moj (commented):
Good stuff. Looks like you've solved the problems with logging we had in the PoC

rambabu-posa (approved):
LGTM

PR #5, opened 2023-03-16
DPR-297: Archive payload to Raw zone.

rambabu-posa:


src/main/java/uk/gov/justice/digital/config/JobParameters.java
```diff
@@ -47,6 +47,10 @@ public Duration getKinesisReaderBatchDuration() {
         return Durations.seconds(parsedDuration);
     }
 
+    public String getRawPath() {
+        return getMandatoryProperty("s3.raw.path");
```
carwynmoj, line 51: If this property is specific to the raw job perhaps `raw.s3.path` is a better way to go here.

src/main/java/uk/gov/justice/digital/job/ArchivePayLoadToRawZoneLoadJob.java
```diff
@@ -0,0 +1,81 @@
...
+            .setPropertyNamingStrategy(PropertyNamingStrategies.KEBAB_CASE)
+            .readerFor(EventRecord.class);
+
+    public static void main(String[] args) {
+        System.out.println("Job started");
+        PicocliRunner.run(ArchivePayLoadToRawZoneLoadJob.class);
+    }
+
+    private static final VoidFunction<JavaRDD<byte[]>> batchProcessor = batch -> {
+        System.out.println("inside batchProcessor.. ");
```
carwynmoj, line 46: Now logging has been merged, consider replacing your `printlns` with logger calls.

src/main/java/uk/gov/justice/digital/job/ArchivePayLoadToRawZoneLoadJob.java
```diff
@@ -0,0 +1,81 @@
...
+    private static final ObjectReader dmsEventReader = new ObjectMapper()
+            .setPropertyNamingStrategy(PropertyNamingStrategies.KEBAB_CASE)
+            .readerFor(EventRecord.class);
+
+    public static void main(String[] args) {
+        System.out.println("Job started");
+        PicocliRunner.run(ArchivePayLoadToRawZoneLoadJob.class);
+    }
+
+    private static final VoidFunction<JavaRDD<byte[]>> batchProcessor = batch -> {
```
carwynmoj, line 45: This method does not need to be static.

src/main/java/uk/gov/justice/digital/job/ArchivePayLoadToRawZoneLoadJob.java
```diff
@@ -0,0 +1,81 @@
...
+        System.out.println("inside batchProcessor.. ");
+
+        val sparkConf = batch.context().getConf();
```
carwynmoj, line 48: Consider moving lines 47 and 48 into a separate method `getSparkSession()` inside this class for now.

src/main/java/uk/gov/justice/digital/job/ArchivePayLoadToRawZoneLoadJob.java
```diff
@@ -0,0 +1,81 @@
...
+
+/**
+ * Test job to read events from a kinesis data stream and writing to S3 raw zone.
+ */
+@Singleton
+@Command(name = "ArchivePayLoadToRawZoneLoadJob")
+public class ArchivePayLoadToRawZoneLoadJob implements Runnable {
+
+    private final KinesisReader kinesisReader;
+    private static RawZone rawZone = null;
```
carwynmoj, line 29: This should not be static. Change to `private final`.

src/main/java/uk/gov/justice/digital/job/ArchivePayLoadToRawZoneLoadJob.java
```diff
@@ -0,0 +1,81 @@
...
+                .set("spark.sql.legacy.charVarcharAsString", "true");
+
+        SparkSession spark = SparkSession.builder()
+                .config(sparkConf)
+                .getOrCreate();
+
+        JavaRDD<Row> rowRDD = batch.map((Function<byte[], Row>) msg -> {
+            String data = new String(msg, StandardCharsets.UTF_8).replace("(\r\n|\n|\r|\t|\\)", "");
+            Row row = RowFactory.create(data);
+            return row;
```
carwynmoj, line 62: `row` isn't adding anything so just return the result `RowFactory.create()`

src/main/java/uk/gov/justice/digital/job/ArchivePayLoadToRawZoneLoadJob.java
```diff
@@ -0,0 +1,81 @@
...
+        });
+
+        System.out.println("rowRDD.isEmpty() ==> " + rowRDD.isEmpty());
+
+        Dataset<Row> raw_df = rawZone.process(rowRDD, spark);
```
carwynmoj, line 67: We're not doing anything with the `raw_df` variable. `process` should probably be `void` and returning nothing.

Also this variable doesn't follow java naming conventions. Should be camelcase starting with lowercase e.g. `rawDataFrame`.

src/main/java/uk/gov/justice/digital/util/SourceReference.java
```diff
@@ -0,0 +1,22 @@
+package uk.gov.justice.digital.util;
```
carwynmoj, line 1: We have a config package so this should probably live there.

src/main/java/uk/gov/justice/digital/job/ArchivePayLoadToRawZoneLoadJob.java
dominic-messenger-moj, line 1: - System.out.println replaced with logger.info()
- Can RDDs be empty ? If so, why are they. Do we know ?
- #26 Should be LoadJob or DataHubJob (not ArchivePayLoadToRawZoneJob )
- #36 static ObjectReader - should be injected
- #49 configuring the session for each batch is overhead . It may also not work - some configurations have to be set before the session is created. This should be set once at the Job Level triggered from main(..)
- #60 is this necessary ?
- #77 do we know - should we continue to run the job in a loop unless there is a formal close 
- Is there a job.commit() requirement ?

src/main/java/uk/gov/justice/digital/zone/RawZone.java
```diff
@@ -0,0 +1,75 @@
...
+
+import javax.inject.Inject;
+import javax.inject.Singleton;
+import java.util.List;
+
+import static org.apache.spark.sql.functions.*;
+import static org.apache.spark.sql.functions.col;
+
+@Singleton
+public class RawZone implements Zone {
```
carwynmoj, line 21: I'm not sure we need to lift the zone concept from the POC code.

This looks like it provides a row processor which you could handle differently e.g by providing job specific code in `job.archivepayloadtoraw`.

This might be something to discuss as a team. My feeling is that it's a little early to start abstracting things like this.

src/main/java/uk/gov/justice/digital/util/SourceReference.java
dominic-messenger-moj, line 1: - FileInputStream is not going to work in the job as the properties are a resource in a jar. You need to read as a resource. See https://github.com/ministryofjustice/digital-prison-reporting-poc/blob/7683b55a546931ba7b92a28e9bdeb416b3a079d1/digital-prisons-reporting/common/src/main/java/uk/gov/justice/dpr/service/SourceReferenceService.java#L90
- Should translate both SOURCE and TABLE names.

src/main/java/uk/gov/justice/digital/zone/RawZone.java
```diff
@@ -0,0 +1,75 @@
...
+    private static String DELTA_FORMAT = "delta";
+    private String rawPath;
+
+    @Inject
+    public RawZone(JobParameters jobParameters){
+        this.rawPath = jobParameters.getRawPath();
+    }
+
+    @Override
+    public Dataset<Row> process(JavaRDD<Row> rowRDD, SparkSession spark) {
```
carwynmoj, line 32: Consider making this void if we're not using the result.

src/main/java/uk/gov/justice/digital/zone/RawZone.java
dominic-messenger-moj, line 1: - System.out.println replaced with logger.info()/warn() etc
- #17 replace with specific functions so we understand the usage
- #33 should probably accept a dataframe rather than an RDD. The DataFrame should already be a DMS record
- #42 this is DMS 3.4.6 specific code. Please extract and put into a utility class and injected into the Job not the Zone. We should create a DMS dataframe BEFORE this, so that we can re-use for structured and curated. This is so we can upgrade to DMS 3.4.7 by simply changing the injected utility instance. This is combining the dataframing of the RDD into a DMS message with archiving to Raw
- #64 should filter on source AND table. There is no guarantee of uniqueness in table names.

**Kinesis -- RDD --> HubJob -- DMS_DF --> RawZone**

src/main/java/uk/gov/justice/digital/zone/Zone.java
dominic-messenger-moj, line 1: #13 should probably accept a dataframe rather than an RDD.

dominic-messenger-moj (commented):
- Missing Job Tests
- Missing RawZone Tests

src/main/java/uk/gov/justice/digital/zone/RawZone.java
rambabu-posa: yes Im about to merge those changes from main branch

carwynmoj (commented):
Added a few comments with a number of suggestions.

We should not use properties files for configuration. These are probably best represented as job parameters or even hardcoded for now until they need to be factored out. 

Take a look and let me know what you think.

src/main/java/uk/gov/justice/digital/zone/RawZone.java
rambabu-posa: I think that Raw DataFrame is going to use by next step like Structured, Curated.

src/main/java/uk/gov/justice/digital/util/SourceReference.java
rambabu-posa: If table names have upper case issue, I fixed it in that dataframe itself. Please suggest me on this if there is any other reason for mapping

dominic-messenger-moj: - some tables aren't tables but materialised views and are named as things like prison_location_1002_mv in NOMIS but should be prison_location in our space.
- some tables need translation from other languages (eg SAP has tables with names like LFM2 which mean nothing)
- some table names contain reserved characters (.$-) that need translation.
- part of the responsibility of the data stewards is to make tables human understandable. Being able to change table names from technical

src/main/java/uk/gov/justice/digital/zone/RawZone.java
dominic-messenger-moj: Agree - probably not needed yet.

src/main/java/uk/gov/justice/digital/util/SourceReference.java
rambabu-posa: Can I reuse the same logic because we are going to find best approach for all these mappings?

src/main/java/uk/gov/justice/digital/job/ArchivePayLoadToRawZoneLoadJob.java
dominic-messenger-moj: If you place the DataFraming in the main job, then the process does not need to return anything. It should just 'write'

carwynmoj (approved):
Thanks for making the requested changes. 👍

PR #6, opened 2023-03-16
DPR-319 remove spark-core from kinesis-asl dep reducing jar size

carwynmoj:
Summary of changes
* Glue already adds Spark 3.3.0 to the classpath so we can remove spark-core from the jar which is pulled in transitively by the kinesis asl streaming library
* This reduces the jar size significantly from approx 160MiB to 36MiB


rambabu-posa (approved):
LGTM

PR #8, opened 2023-03-17
DPR-297 Raw Zone unit tests.

rambabu-posa:


src/main/java/uk/gov/justice/digital/service/SourceReferenceService.java
```diff
@@ -39,6 +39,10 @@ public class SourceReferenceService {
             REF.put("oms_owner.agency_internal_locations", new SourceReference("SYSTEM.AGENCY_INTERNAL_LOCATIONS", "nomis", "agency_internal_locations", "INTERNAL_LOCATION_ID", getSchemaFromResource("/schemas/oms_owner.agency_internal_locations.schema.json")));
 
 
+            // aurora postgres
```
carwynmoj, line 42: We can skip these entries for now so you can remove these.

carwynmoj (commented):
Just a quick comment re. the sources. It's OMS only so no need to add the aurora postgres ones.

src/main/java/uk/gov/justice/digital/service/SourceReferenceService.java
carwynmoj: Useful for some internal testing so leave these in.

carwynmoj (approved):
LGTM 👍

PR #17, opened 2023-03-30
DPR-365 ensure delta metadata is updated after appending data

carwynmoj:
Summary of changes
* ensure we update metadata after appending data to a given delta table
* other minor tidy ups including
  * fixing potential NPE in CuratedZone
  * removing unused imports etc..


dominic-messenger-moj (commented):
LGTM

PR #18, opened 2023-04-03
DPR-368 filter DMS control records

carwynmoj:
Summary of changes
* revise the parsing logic so we allow the JSON to parse in permissive mode so that we can filter control records which would otherwise fail this stage since they have no data property
* once filtered we then construct the data frame so that it matches the parsed data schema we use internally at which point the schema can be strictly applied to detect any structural issues with the data records that we are interested in


rambabu-posa (approved):
LGTM

PR #20, opened 2023-04-14
NOJIRA add helper scripts for working with aws glue jobs

carwynmoj:
Summary of changes

Added the following scripts
 * push-jar - builds and pushes the project jar to an s3 folder uri
 * tail-glue-logs - tail the logs for a given job name
 * start-and-tail-job - starts a given job name and then tails the logs

### DIFF
--- a/scripts/push-jar
+++ b/scripts/push-jar
@@ -1,0 +1,36 @@
+#!/bin/bash
+
+# Script to build the jobs jar and push it to a specified s3 bucket.
+#
+# Usage:
+#
+#   scripts/push-jar <s3-folder-uri>
+#
+#   e.g.
+#
+#   scripts/push-jar s3://dpr-jars/dev
+
+set -e
+
+s3_uri="$@"
+
+echo
+
+if [ -z "$s3_uri" ]
+then
+  echo "A destination s3 uri must be specified"
+  exit 1;
+fi
+
+echo "Building jar"
+echo
+
+script_dir="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+
+./gradlew clean shadowJar
+
+echo
+echo "Uploading jar to $s3_uri"
+echo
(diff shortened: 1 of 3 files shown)


scripts/push-jar
```diff
@@ -0,0 +1,36 @@
...
+echo
+
+script_dir="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+
+./gradlew clean shadowJar
+
+echo
+echo "Uploading jar to $s3_uri"
+echo
+aws s3 cp ./build/libs/digital-prison-reporting-jobs-1.0-SNAPSHOT-all.jar $s3_uri/digital-prison-reporting-jobs-1.0-SNAPSHOT-all.jar
```
kartthik-n, line 35: can you please make the version 1.0-SNAPSHOT as variable? this will change as per the release

carwynmoj: Ahh yeah, good spot. Afaik we're not doing this dynamically yet but I'll revise this to match on the static parts of the jar name instead.

carwynmoj: Pushed up changes now which also includes a fix to ensure the gradle build is run in the correct directory.

scripts/push-jar
```diff
@@ -0,0 +1,37 @@
...
+echo "Building jar"
+echo
+
+cd $script_dir/../
+./gradlew clean shadowJar
+
+echo
+echo "Uploading jar to $s3_uri"
+echo
+aws s3 cp ./build/libs/digital-prison-reporting-jobs-*-all.jar $s3_uri/digital-prison-reporting-jobs-1.0-SNAPSHOT-all.jar
```
kartthik-n, line 36: do we need to specify the destination file name here? can we simplify like this?

aws s3 cp ./build/libs/digital-prison-reporting-jobs-*-all.jar $s3_uri/

carwynmoj: You're quite right. This is a hangover from a version I knocked up for myself a while back. Should've read the whole line.

carwynmoj: Fix pushed for this (dest filename has been removed) so resolving this.

kartthik-n (approved):

PR #33, opened 2023-04-28
DPR-408 enable gradle caching

carwynmoj:
Summary of changes
* enable gradle caching
* enable gradle parallelism for add gradle tasks (previously was test only)

Note
* gradle caching will cache the result of successful operations so they're only run once
* for example running the tests twice, the second attempt will complete immediately because the first run passed
* if you need to force a task to run for whatever reason just add the `--rerun-tasks` option to your gradle command

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -1,0 +1,2 @@
+org.gradle.parallel=true
+org.gradle.caching=true


carwynmoj:
Closing for now since this doesn't appear to have the desired effect on circle ci which is a shame. :(

PR #34, opened 2023-04-30
Dpr 219 domain schemas

kartthik-n:


src/main/java/uk/gov/justice/digital/domain/DomainExecutor.java
dominic-messenger-moj, line 1: 1. DomainSchemaService  should be injected
2. Line 81-92 should really be in the DomainSchemaService::createIfNotExists(info)
3. Creation should be done at Line 76
4. Same for Line 102-113 - code should be encapsulated in DomainSchemaService
5. Same for 149-158

carwynellis:
Took a quick look and had the following observations.

1. `getGlueClient` should be removed from the `JobClient` and `JobParameters` classes and replaced with a new client, like `JobClient` that wraps (and hides) the implementation details of the AWS client. If it's just needed to create a hive database then consider calling it `HiveDatabaseClient` and then *inject* it into the classes that need it. The point of the client layer is to hide platform specific implementation detail and provide a simpler API for the code using it (plus you will then have a `@Singleton` that can be injected where it's needed)
2. DomainExecutor line 102 on. The double if should be collapsed into a single if combining the tests with &&
3. `log.error` calls should *always* take the form `log.error("Description", exception)` in order for the stacktrace and other useful information to be logged - please fix any log.error calls that do not follow this form
4. Remove test specific code from the DomainSchemaService - no test code should be deployed to production - move into tests
5. In some of your tests you have some `static` fields that are initially set to `null` with a `@BeforeAll` setting them. This isn't necessary. Just set the `static` fields directly where declared.
6. Your circle CI build is failing because your tests are using a real AWSGlue client.  Test code should *not* be making requests to real services for a variety of reasons, including reliability. Replace any collaborators that your tests need with mocks. Micronaut makes this easy. Take a look at https://github.com/ministryofjustice/digital-prison-reporting-domain-builder/blob/main/backend/src/test/kotlin/uk/gov/justice/digital/controller/DomainControllerTest.kt which demonstrates the use of the `@MicronautTest` and `@MockBean` annotations along with code demonstrating how to set up mock responses.

Haris-Dilgenta (approved):

PR #40, opened 2023-05-05
DPR-374:Refactoring changes taken from DPR-418

kartthik-n:
DPR-418 changes included in DPR-374 now


dominic-messenger-moj (approved):
This is approved.
There are some integration tests posing as unit tests that cause intermittent failures (the hive metastore db running locally) that will need shoring up as part of DPR-420.
However, this builds and any further changes required (such as protecting public methods etc) can be and should be done as part of DPR-420

PR #41, opened 2023-05-09
DPR-422 address remaining code quality issues

carwynmoj:
Summary of changes
* introduce `ClientProvider<T>` interface and provide implementation for `AWSGlue` client instance
* revise `JobClient` to use `GlueClientProvider` to obtain an instance of the `AWSGlue` client
* revise `DomainSchemaService` to use `GlueClientProvider`
* revise `Properties` class, now a singleton renamed to `JobProperties` with classes using this now injecting in an instance
* revise tests to provide mock collaborators (in this case for ClientProvider and underlying AWSGlue client instance along with the JobProperties singleton)
* revise `DomainSchemaService` tests to use mocked collaborators including mocked `Dataframe` and `Schema` instances so the test no longer depends on a running spark instance
* other minor changes including some tidying, rename of `saveToDisk` to `persistDataset`

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -26,6 +26,7 @@ ext {
     junitVersion = '5.8.1'
     log4jVersion = '2.20.0'
     micronautVersion = '3.8.7'
+    mockitoVersion = '4.11.0'
     sparkVersion = '3.3.0'
 }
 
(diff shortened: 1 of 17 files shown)


dominic-messenger-moj (commented):
Looks great. Thank you!

dominic-messenger-moj (commented):
Looks great. Thank you!

PR #44, opened 2023-05-11
DPR-422 skip tests on build since they have already been run

carwynmoj:


carwynmoj:
Closing this PR since the attempts to speed up the tests yielded marginal gains for additional complexity such that the latter isn't really worth it. A new branch will be created without the speed up commits and pushed as a PR after some quick tidying.

PR #45, opened 2023-05-11
DPR-437 Combined JobParameters with Job Arguments

dominic-messenger-moj:


carwynmoj:
Do we need this? If you're just using this to log out the parameters glue already does this at the point the jar is launched.

carwynmoj:
> Do we need this? If you're just using this to log out the parameters glue already does this at the point the jar is launched.

Discussed with Dom and realised I'd missed that instead of using the client we're now taking the args direct from main which is an improvement.

carwynmoj (approved):
👍 LGTM

PR #53, opened 2023-05-17
DPR-464 Fix insert and delete errors in creating hive metadata tables.

dominic-messenger-moj:
See scenarios in DPR-464

### DIFF
--- a/src/main/java/uk/gov/justice/digital/domain/DomainExecutor.java
+++ b/src/main/java/uk/gov/justice/digital/domain/DomainExecutor.java
@@ -60,13 +60,21 @@ public class DomainExecutor {
         String tablePath = storage.getTablePath(info.getBasePath(), info.getSchema(), info.getTable());
         logger.info("Domain insert to disk started");
         try {
-            if (!storage.exists(spark, info)) {
-                storage.create(tablePath, dataFrame);
-                schema.create(info, tablePath, dataFrame);
-                logger.info("Creating delta table completed...");
+            if(storage.exists(spark, info)) {
+                if(storage.hasRecords(spark, info)) {
+                    // this is trying to overwrite so throw an exception
+                    throw new DomainExecutorException("Delta table " + info.getTable() + " already exists and has records. Try replace");
+                } else {
+                    storage.replace(tablePath, dataFrame);
+                    logger.warn("Storage is empty so performing a replace instead");
+                }
             } else {
-                throw new DomainExecutorException("Delta table " + info.getTable() + "already exists");
+                storage.create(tablePath, dataFrame);
             }
+            // create the schema
+            schema.create(info, tablePath, dataFrame);
+            logger.info("Creating delta table completed...");
+
         } catch (DomainSchemaException | DataStorageException dse) {
             throw new DomainExecutorException(dse);
         }
@@ -106,12 +114,12 @@ public class DomainExecutor {
         try {
             if (storage.exists(spark, info)) {
                 storage.delete(spark, info);
-                schema.drop(info);
                 storage.endTableUpdates(spark, info);
             } else {
-                throw new DomainExecutorException("Table " + info.getSchema() + "." + info.getTable() +
-                        " doesn't exists");
+                logger.warn("Delete table " + info.getSchema() + "." + info.getTable() + " not executed as table doesn't exist");
             }
+
+            schema.drop(info);
         } catch (DomainSchemaException | DataStorageException dse) {
             throw new DomainExecutorException(dse);
         }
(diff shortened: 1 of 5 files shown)


src/main/java/uk/gov/justice/digital/domain/DomainExecutor.java
```diff
@@ -106,12 +114,12 @@ protected void deleteTable(TableIdentifier info) throws DomainExecutorException
         try {
             if (storage.exists(spark, info)) {
                 storage.delete(spark, info);
-                schema.drop(info);
                 storage.endTableUpdates(spark, info);
             } else {
-                throw new DomainExecutorException("Table " + info.getSchema() + "." + info.getTable() +
-                        " doesn't exists");
+                logger.warn("Delete table " + info.getSchema() + "." + info.getTable() + " not executed as table doesn't exist");
```
carwynmoj, line 119: No change needed, just a comment ICYMI that logger calls can also take a simple format string e.g. `logger.warn("Got {} Expected {}", "foo", "bar");`

src/main/java/uk/gov/justice/digital/service/DataStorageService.java
```diff
@@ -6,6 +6,7 @@
 import org.apache.spark.sql.SparkSession;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+import scala.Array;
```
carwynmoj, line 9: Do we need this import?

src/main/java/uk/gov/justice/digital/service/DataStorageService.java
```diff
@@ -34,6 +35,16 @@ public String getTablePath(String... elements) {
         return String.join("/", elements);
     }
 
+    public boolean hasRecords(final SparkSession spark, final TableIdentifier info) throws DataStorageException {
+        logger.info("Checking details for Delta table..." + info.getSchema() + "." + info.getTable());
+        if(exists(spark, info)) {
+            Dataset<Row> df = load(spark, info);
+            return !df.isEmpty();
+        } else {
+            return false;
```
carwynmoj, line 44: Just checking that it's ok to return false here when the table doesn't exist?

src/test/java/uk/gov/justice/digital/service/DataStorageServiceTest.java
```diff
@@ -64,6 +64,54 @@ void shouldReturnFalseWhenStorageDoesNotExists() {
...
+        TableIdentifier info = new TableIdentifier("s3://test-bucket", "domain",
+                "incident", "demographics");
+        String identifier = testStorage.getTablePath(info.getBasePath(), info.getSchema(), info.getTable());
+        when(DeltaTable.isDeltaTable(spark, identifier)).thenReturn(true);
+        boolean actualResult = testStorage.exists(spark, info);
+        assertTrue(actualResult);
+
+        when(DeltaTable.forPath(spark, identifier)).thenReturn(mockedDeltaTable);
+        when(mockedDeltaTable.toDF()).thenReturn(df);
+        boolean hasRecords = testStorage.hasRecords(spark, info);
```
carwynmoj, line 80: minor - I'd probably inline this so you have `assertTrue(testStorage.....)` but NABD.

carwynmoj (approved):
LGTM 👍 

Just a few minor remarks that you might want to look at.

src/main/java/uk/gov/justice/digital/service/DataStorageService.java
carwynmoj: Reviewed the ACs in DPR-464 which confirms that this is fine.

src/main/java/uk/gov/justice/digital/service/DataStorageService.java
dominic-messenger-moj: No. Removed

src/main/java/uk/gov/justice/digital/domain/DomainExecutor.java
```diff
@@ -60,13 +60,21 @@ protected void insertTable(TableIdentifier info, Dataset<Row> dataFrame)
...
+                if(storage.hasRecords(spark, info)) {
+                    // this is trying to overwrite so throw an exception
+                    throw new DomainExecutorException("Delta table " + info.getTable() + " already exists and has records. Try replace");
+                } else {
+                    storage.replace(tablePath, dataFrame);
+                    logger.warn("Storage is empty so performing a replace instead");
+                }
             } else {
-                throw new DomainExecutorException("Delta table " + info.getTable() + "already exists");
+                storage.create(tablePath, dataFrame);
```
kartthik-n, line 72: create method not having an overwrite (.option("overwriteSchema", true))option. Can we test this scenario to ensure it works without the overwrite option. Instead we can call update() method in this block

PR #54, opened 2023-05-17
DPR-444 ensure path generation produces valid paths plus some refactoring

carwynmoj:
Summary of changes
* use `URI` to build and `normalize` paths
* factor out code that builds paths into a static method (this will be used elsewhere in the next PR)
* move `Columns` into `common` package and rename to `ColumnNames`
* refactor Zone tests removing duplication, unnecessary use of spark and redundant test cases where relevant
* other minor changes including removal of dead code

Note - a second PR will follow with another simpler refactor.


src/main/java/uk/gov/justice/digital/common/ColumnNames.java
dominic-messenger-moj, line 1: These 'column names' relate to DMS and are not common as such. I suggest this is moved into the dms package.

dominic-messenger-moj (commented):
Columns relates to DMS and not common. If we replace DMS with Debezium or GoldenGate then each of those will have it's own column structure.
This has leaked out into our code. We should have a 'payload', 'operation', 'timestamp', source etc which will be common across implementations but not what we have now.

src/main/java/uk/gov/justice/digital/common/ColumnNames.java
carwynmoj: Will do this in the next PR.

src/main/java/uk/gov/justice/digital/domain/DomainExecutor.java
```diff
@@ -57,7 +59,7 @@ public DomainExecutor(JobArguments jobParameters,
     protected void insertTable(TableIdentifier info, Dataset<Row> dataFrame)
             throws DomainExecutorException {
         logger.info("DomainExecutor:: insertTable");
-        String tablePath = storage.getTablePath(info.getBasePath(), info.getSchema(), info.getTable());
+        String tablePath = createValidatedPath(info.getBasePath(), info.getSchema(), info.getTable());
```
kartthik-n, line 62: are we catching the exception IllegalArgumentException here?

carwynmoj: The intent is to fail fast here. I think this will be fine but we should see how it plays on test but I don't expect there will be any issues here.

kartthik-n (approved):

PR #56, opened 2023-05-18
Dpr 420 refactoring continued

kartthik-n:


src/main/java/uk/gov/justice/digital/domain/DomainExecutor.java
dominic-messenger-moj, line 1: I am not happy with this Exception.
The underscore only makes sense for the SCHEMA but not the STORAGE which is throwing the exception.
It should read tablePath

dominic-messenger-moj (commented):
Exception is wrong. Should relate to DeltaTable

carwynmoj (changes requested):
LGTM but please just revert your changes to the KinesisReader since I think they were only made to support the test? Otherwise all good although there's a comment from Dom you might want to look at too.

dominic-messenger-moj (commented):
All good

carwynmoj (approved):
LGTM 👍

PR #57, opened 2023-05-18
DPR-444 add toPath() method to TableInfo and move ColumnNames into more specific locations

carwynmoj:
Summary of changes
* added `TableIdentifier.toPath()` as a convenience which cleans up some noise in the code where a path is generated from a `TableIdentifer`
* as requested in the previous PR, removed `ColumnNames` and moved the fields closer to where they're being used
* the `updateManifest()` method on `DataStorageService` no-longer swallows any exceptions being thrown which may highlight hidden issues on test
  * the tests now use an in memory derby DB since this method formerly hid an error on multiple spark instances attempting to use the same derby db which only permits one thread at a time)
* some minor tidying

Note - I've left a TODO in the Converter class. At some point we'll maybe need to look at defining input/output formats more formally. I think for now it's fine as is and it's probably better to wait until we start handling another format before making changes so we can be sure that any design decisions actually make sense.


src/test/java/uk/gov/justice/digital/zone/Fixtures.java
```diff
@@ -16,15 +15,15 @@ public class Fixtures {
 
     public static final String TABLE_SOURCE = "oms_owner";
     public static final String TABLE_NAME = "agency_internal_locations";
-    public static final String OPERATION = "load";
+    public static final String TABLE_OPERATION = "load";
```
kartthik-n, line 18: one general comment. Can we move this Fixtures.java to test.helpers folder?

carwynmoj: The idea is that they're specific to the zone tests which is why they are in this location. Also they're fixtures rather than helpers in this instance.

dominic-messenger-moj (approved):
LGTM

PR #58, opened 2023-05-19
DPR-476 JSON schema defining a dpr domain contract

carwynmoj:
Summary of changes
* added a test that applies the JSON schema to a sample contract to verify that it passes
* the JSON schema supports
  * database top level fields
  * table definitions
  * simple fields
  * enum fields (unlikely we will need other complex type but they're easy to add in future if needed)
  * default values where explicitly stated
 
Notes
* the `validation` field is not implemented yet since it's not entirely clear what this represents
* validation of fields with a format such as the retention periods is not implemented - this could be achieved with regex based rules if needed


dominic-messenger-moj (approved):
LGTM

PR #59, opened 2023-05-22
NOJIRA - reinstate info level logging

carwynmoj:
Summary of changes
* explicitly set log level to INFO
* use micronaut LogLevel enum instead of local strings
* fix naming of `handleInvalidRecords` method


kartthik-n (approved):
Looks Good

PR #60, opened 2023-05-22
DPR-478 some refactoring before working on the SourceReferenceService

carwynmoj:
Summary of changes
* introduce separate `doDomainDelete()` method on the executor so we have a clear separation of this functionality with a cleaner API that no-longer means we need to pass nulls in the parameter list
* internalised the dynamo db table name so that the caller is not required to pass this in anymore - this is an internal concern of the client and shouldn't be exposed
* revised the DomainService tests so they actually test the DomainService (not the DomainExecutor which is already tested elsewhere) using mocked collaborators
* revised the public API of the DomainDefinitionClient so the caller simply makes one method call to get a domain definition (getting the result and parsing it are both internal concerns of the client)
* log messages now use consistent language in Zone classes
* other minor tidying


dominic-messenger-moj (commented):
LGTM

PR #62, opened 2023-05-23
DPR-475: Schema for offenders added

kartthik-n:


src/test/resources/contracts/offenders.avsc
```diff
@@ -0,0 +1,465 @@
...
+  "database": {
+    "type": "oracle",
+    "version": "12",
+    "name": "CNOMT3"
+  },
+  "propagation-policy": "1.day",
+  "retention-policy": "999.years",
+  "fields" : [
+    {
+      "name": "OFFENDER_ID",
```
carwynmoj, line 22: Just a general note. Where we've declared a default value in the schema that defines our format, you can omit the value. In this case this would apply to
* `sensitivity` - default value declared as `public`
* `certified` - default value declared as `false`

carwynmoj (approved):
Looks fine to me.

Just a general observation regarding fields that have a default value. Where your definition matches the default these could be removed since they're effectively redundant.

src/test/resources/contracts/offenders.avsc
kartthik-n: I will leave it to @dominic-messenger-moj to comment on this since the wiki page says the default value for the above. Happy to remove if we all agreed in a tiny PR. Merging this now

dominic-messenger-moj: Propagation Policy should be empty (by default propagation is incremental)
Retention Policy should be blank (by default this is the departmental policy)

dominic-messenger-moj: Sensitivty and Certified can be removed for all fields.

dominic-messenger-moj: Also remove the glossary-index for now for all fields except for offender:id. The remainder we don't actually know what they mean and we shouldn't make any assumptions.

dominic-messenger-moj (commented):
Some minor removals (glossary-index for most, sensitivity and certified for all)
Otherwise all good

dominic-messenger-moj (commented):
Some minor removals (glossary-index for most, sensitivity and certified for all)
Otherwise all good

PR #68, opened 2023-05-26
DPR-475: Schema for offenders modified

kartthik-n:


src/test/resources/contracts/offenders.avsc
```diff
@@ -15,72 +15,52 @@
     "version": "12",
     "name": "CNOMT3"
   },
-  "propagation-policy": "1.day",
-  "retention-policy": "999.years",
+  "propagation-policy": "",
```
carwynmoj, line 18: We could omit these fields for now since I think we need to figure out exactly what these will contain.

carwynmoj (approved):
LGTM 👍 

Minor comment re. the policy fields which could be removed until we have a clear definition for them.

src/test/resources/contracts/offenders.avsc
kartthik-n: @dominic-messenger-moj mentioned to leave it blank. will remove this based on your comment now

PR #69, opened 2023-05-26
DPR-478 integrate with glue schema registry

carwynmoj:
Summary of changes
* introduce GlueSchemaClient and tests
* add new job argument `dpr.contract.registryName`

### DIFF
--- a/src/main/java/uk/gov/justice/digital/client/glue/GlueSchemaClient.java
+++ b/src/main/java/uk/gov/justice/digital/client/glue/GlueSchemaClient.java
@@ -1,0 +1,25 @@
+package uk.gov.justice.digital.client.glue;
+
+import com.amazonaws.services.glue.AWSGlue;
+import jakarta.inject.Inject;
+import jakarta.inject.Singleton;
+import uk.gov.justice.digital.config.JobArguments;
+
+@Singleton
+public class GlueSchemaClient {
+
+    private final AWSGlue glueClient;
+    private final JobArguments jobArguments;
+
+    @Inject
+    public GlueSchemaClient(GlueClientProvider glueClientProvider,
+                            JobArguments jobArguments) {
+       this.glueClient = glueClientProvider.getClient();
+       this.jobArguments = jobArguments;
+    }
+
+    // TODO - determine what we need to pass in
+    public void getSchema() {
+
+    }
+}
--- a/src/main/java/uk/gov/justice/digital/client/glue/GlueSchemaClient.java
+++ b/src/main/java/uk/gov/justice/digital/client/glue/GlueSchemaClient.java
@@ -1,25 +1,52 @@
 package uk.gov.justice.digital.client.glue;
 
 import com.amazonaws.services.glue.AWSGlue;
+import com.amazonaws.services.glue.model.EntityNotFoundException;
+import com.amazonaws.services.glue.model.GetSchemaVersionRequest;
+import com.amazonaws.services.glue.model.SchemaId;
+import com.amazonaws.services.glue.model.SchemaVersionNumber;
 import jakarta.inject.Inject;
 import jakarta.inject.Singleton;
+import lombok.val;
 import uk.gov.justice.digital.config.JobArguments;
+
+import java.util.Optional;
 
 @Singleton
 public class GlueSchemaClient {
 
     private final AWSGlue glueClient;
-    private final JobArguments jobArguments;
+    private final String contractRegistryName;
 
     @Inject
     public GlueSchemaClient(GlueClientProvider glueClientProvider,
                             JobArguments jobArguments) {
-       this.glueClient = glueClientProvider.getClient();
-       this.jobArguments = jobArguments;
+        this.glueClient = glueClientProvider.getClient();
+        this.contractRegistryName = jobArguments.getContractRegistryName();
     }
 
-    // TODO - determine what we need to pass in
-    public void getSchema() {
-
+    public Optional<String> getSchema(String schemaName) {
+        try {
+            val result = glueClient.getSchemaVersion(createRequestForSchemaName(schemaName));
+            val schemaData = result.getSchemaDefinition();
+            return Optional.of(schemaData);
+        }
+        catch (EntityNotFoundException e) {
+            return Optional.empty();
+        }
     }
+
+    private GetSchemaVersionRequest createRequestForSchemaName(String schemaName) {
+        val schemaId = new SchemaId()
+                .withRegistryName(contractRegistryName)
+                .withSchemaName(schemaName);
+
+        val latestSchemaVersion = new SchemaVersionNumber()
+                .withLatestVersion(true);
+
+        return new GetSchemaVersionRequest()
+                .withSchemaId(schemaId)
+                .withSchemaVersionNumber(latestSchemaVersion);
+    }
+
 }
--- a/src/test/java/uk/gov/justice/digital/client/glue/GlueSchemaClientTest.java
+++ b/src/test/java/uk/gov/justice/digital/client/glue/GlueSchemaClientTest.java
@@ -1,0 +1,94 @@
+package uk.gov.justice.digital.client.glue;
+
+import com.amazonaws.SdkClientException;
+import com.amazonaws.services.glue.AWSGlue;
+import com.amazonaws.services.glue.model.EntityNotFoundException;
+import com.amazonaws.services.glue.model.GetSchemaVersionResult;
+import lombok.val;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.ExtendWith;
+import org.mockito.Mock;
+import org.mockito.junit.jupiter.MockitoExtension;
+import uk.gov.justice.digital.config.JobArguments;
+
+import java.util.Optional;
+
+import static org.junit.jupiter.api.Assertions.*;
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.Mockito.when;
+
+@ExtendWith(MockitoExtension.class)
+public class GlueSchemaClientTest {
+
+    private static final String SCHEMA_NAME = "somesource.sometable";
+    private static final String FAKE_SCHEMA_DEFINITION = "This is a fake schema definition";
+
+    @Mock
+    private GlueClientProvider mockClientProvider;
+
+    @Mock
+    private AWSGlue mockClient;
+
+    @Mock
+    private JobArguments mockArguments;
+
+    @Mock
+    private GetSchemaVersionResult mockResponse;
+
+    private GlueSchemaClient underTest;
+
+    @BeforeEach
+    public void setup() {
+        givenJobArgumentsReturnAContractRegistryName();
+        givenClientProviderReturnsAClient();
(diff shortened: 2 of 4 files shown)


src/main/java/uk/gov/justice/digital/client/glue/GlueSchemaClient.java
```diff
@@ -0,0 +1,52 @@
...
+    private final String contractRegistryName;
+
+    @Inject
+    public GlueSchemaClient(GlueClientProvider glueClientProvider,
+                            JobArguments jobArguments) {
+        this.glueClient = glueClientProvider.getClient();
+        this.contractRegistryName = jobArguments.getContractRegistryName();
+    }
+
+    public Optional<String> getSchema(String schemaName) {
```
kartthik-n, line 28: throw the exception to the calling place. Don't catch it in this method

src/main/java/uk/gov/justice/digital/client/glue/GlueSchemaClient.java
```diff
@@ -0,0 +1,52 @@
...
+        try {
+            val result = glueClient.getSchemaVersion(createRequestForSchemaName(schemaName));
+            val schemaData = result.getSchemaDefinition();
+            return Optional.of(schemaData);
+        }
+        catch (EntityNotFoundException e) {
+            return Optional.empty();
```
kartthik-n, line 35: no logger inside catch block

src/main/java/uk/gov/justice/digital/client/glue/GlueSchemaClient.java
```diff
@@ -0,0 +1,52 @@
...
+        }
+    }
+
+    private GetSchemaVersionRequest createRequestForSchemaName(String schemaName) {
```
kartthik-n, line 39: make the modifier as final in the parameter schemaName

src/test/java/uk/gov/justice/digital/client/glue/GlueSchemaClientTest.java
```diff
@@ -0,0 +1,94 @@
...
+    }
+
+    @Test
+    public void shouldThrowAnExceptionIfAnyOtherErrorOccurs() {
+        givenClientThrowsSomeOtherException();
+
+        assertThrows(SdkClientException.class, () -> underTest.getSchema(SCHEMA_NAME));
+    }
+
+    private void givenClientProviderReturnsAClient() {
```
kartthik-n, line 73: this is doing nothing here. remove the test. you are not mocking any behaviour

src/test/java/uk/gov/justice/digital/client/glue/GlueSchemaClientTest.java
```diff
@@ -0,0 +1,94 @@
...
+        when(mockClientProvider.getClient()).thenReturn(mockClient);
+    }
+
+    private void givenJobArgumentsReturnAContractRegistryName() {
```
kartthik-n, line 77: same here

src/test/java/uk/gov/justice/digital/client/glue/GlueSchemaClientTest.java
```diff
@@ -0,0 +1,94 @@
...
+        when(mockArguments.getContractRegistryName()).thenReturn("test-contract-registry");
+    }
+
+    private void givenClientReturnsAValidResponse() {
```
kartthik-n, line 81: same here

src/test/java/uk/gov/justice/digital/client/glue/GlueSchemaClientTest.java
```diff
@@ -0,0 +1,94 @@
...
+        when(mockResponse.getSchemaDefinition()).thenReturn(FAKE_SCHEMA_DEFINITION);
+        when(mockClient.getSchemaVersion(any())).thenReturn(mockResponse);
+    }
+
+    private void givenClientThrowsAnEntityNotFoundException() {
```
kartthik-n, line 86: same here

src/test/java/uk/gov/justice/digital/client/glue/GlueSchemaClientTest.java
```diff
@@ -0,0 +1,94 @@
...
+        when(mockClient.getSchemaVersion(any())).thenThrow(new EntityNotFoundException("Schema not found"));
+    }
+
+    private void givenClientThrowsSomeOtherException() {
```
kartthik-n, line 90: same here

dominic-messenger-moj (commented):
LGTM

dominic-messenger-moj (commented):
LGTM

src/main/java/uk/gov/justice/digital/client/glue/GlueSchemaClient.java
carwynmoj: This exception is specifically thrown in response to a 404 hence returning an Optional.empty.

src/main/java/uk/gov/justice/digital/client/glue/GlueSchemaClient.java
carwynmoj: We agreed that final parameters don't really add anything.

src/test/java/uk/gov/justice/digital/client/glue/GlueSchemaClientTest.java
carwynmoj: This given method sets up required mocked state for the tests.

src/test/java/uk/gov/justice/digital/client/glue/GlueSchemaClientTest.java
carwynmoj: As above.

src/test/java/uk/gov/justice/digital/client/glue/GlueSchemaClientTest.java
carwynmoj: As above.

src/test/java/uk/gov/justice/digital/client/glue/GlueSchemaClientTest.java
carwynmoj: As other remarks - this is setting up mocks for the test.

src/main/java/uk/gov/justice/digital/client/glue/GlueSchemaClient.java
kartthik-n: this is not a correct way of coding. You should call it in the calling place. The method just does what it needs to do, the calling place has to decide the action. We have cutoms exceptions to handle

src/main/java/uk/gov/justice/digital/client/glue/GlueSchemaClient.java
kartthik-n: logger inside the catch block is needed. It has lot of meaning what the exception is for debugging purpose

src/main/java/uk/gov/justice/digital/client/glue/GlueSchemaClient.java
kartthik-n: not necessarily true. The method should not alter the original data it sent. If you doesn't make it final, the method can alter the value and the calling place may have the corrupted value

src/test/java/uk/gov/justice/digital/client/glue/GlueSchemaClientTest.java
kartthik-n: sorry, when doesn't really calling the method. it has no meaning. test it with log statements inside the method and see whether it is getting called. When method itself not called what is the purpose of mocking?

src/test/java/uk/gov/justice/digital/client/glue/GlueSchemaClientTest.java
kartthik-n: same comment as above

src/test/java/uk/gov/justice/digital/client/glue/GlueSchemaClientTest.java
kartthik-n: same comment as above

src/test/java/uk/gov/justice/digital/client/glue/GlueSchemaClientTest.java
kartthik-n: same comment as above

src/main/java/uk/gov/justice/digital/client/glue/GlueSchemaClient.java
carwynmoj: Not in the 404 case. The AWS client throws a specific exception which the caller doesn't have to deal with. This results in cleaner code at the call site.

src/main/java/uk/gov/justice/digital/client/glue/GlueSchemaClient.java
carwynmoj: Not necessarily. The caller should decide whether it's an error or not. In this case, the source reference service will fallback to the static resources that we have now and log. Later when we remove the fallbacks, it will throw an exception.

src/main/java/uk/gov/justice/digital/client/glue/GlueSchemaClient.java
carwynmoj: This is extremely unlikely and when one considers the cost / benefit it's not really worth the added visual noise.

src/test/java/uk/gov/justice/digital/client/glue/GlueSchemaClientTest.java
carwynmoj: The `getClient()` method is being called in line 24 of the GlueSchemaClient. Hence this mock is necessary so that we return a mocked instance of the `AWSGlue` class.

src/test/java/uk/gov/justice/digital/client/glue/GlueSchemaClientTest.java
carwynmoj: This mock is necessary since we need the arguments to return something when the glue client calls the `getContractRegistry()` name. See GlueSchemaClient line 25.

src/test/java/uk/gov/justice/digital/client/glue/GlueSchemaClientTest.java
carwynmoj: This mock specifically tests the handling of the `EntityNotFoundException` which is thrown by `AWSGlue` when no schema is found. We use this mock to confirm that the client returns an empty optional, which is the expected behaviour for this exception.

src/test/java/uk/gov/justice/digital/client/glue/GlueSchemaClientTest.java
carwynmoj: This mock is used to confirm that any other exception thrown by the underlying AWS client is propagated to the caller. Typically in these cases there isn't much we can do so the caller needs to decide whether to catch or propagate.

carwynmoj:
Will do ahead any merge this PR on the basis that it can be easily reverted if there are any strong objections. For reference, the client is fully tested and has 100% code coverage.

PR #70, opened 2023-05-26
DPR-478 add avro to spark schema converter

carwynmoj:
Summary of changes
* added converter that takes an avro schema and converts it into a spark schema using the spark-avro support
* added tests to cover conversion of all declared types
* simplified converter so we have an interface with a single method and refactored

### DIFF
--- a/src/main/java/uk/gov/justice/digital/converter/Converter.java
+++ b/src/main/java/uk/gov/justice/digital/converter/Converter.java
@@ -1,73 +1,7 @@
 package uk.gov.justice.digital.converter;
 
-import org.apache.spark.api.java.JavaRDD;
-import org.apache.spark.sql.Dataset;
-import org.apache.spark.sql.Row;
-import org.apache.spark.sql.SparkSession;
-import org.apache.spark.sql.types.StructType;
+public interface Converter<I, O> {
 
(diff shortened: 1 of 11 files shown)


src/main/java/uk/gov/justice/digital/converter/Converter.java
dominic-messenger-moj, line 1: The original Converter is a MessageConverter to convert CDC messages from Oracle GoldenGate, AWS DMS, Debezium and others into a common internal format.

This converter is a generic 'Converter' that converts a generic thing to another.
I suggest you move this to a MessageConverter and NOT to DMS_3_4_6 (which is just one type of message converter)

src/main/java/uk/gov/justice/digital/converter/dms/DMS_3_4_6.java
dominic-messenger-moj, line 1: As elsewhere, the Converter should be renamed to MessageConverter (as a base class for ALL message converters)

carwynmoj: As discussed I'll get something on the backlog for this and we can bring back a common base class when we start implementing another dms format.

PR #71, opened 2023-05-29
DPR-480 + DPR-489 Improved raw storage and de-duped raw, structured a…

dominic-messenger-moj:
…nd curated zones

### DIFF
--- a/src/main/java/uk/gov/justice/digital/converter/Converter.java
+++ b/src/main/java/uk/gov/justice/digital/converter/Converter.java
@@ -6,32 +6,67 @@ import org.apache.spark.sql.Row;
 import org.apache.spark.sql.SparkSession;
 import org.apache.spark.sql.types.StructType;
 
+import java.util.Arrays;
+import java.util.Optional;
+
 import static org.apache.spark.sql.types.DataTypes.StringType;
 import static uk.gov.justice.digital.converter.Converter.ParsedDataFields.*;
 
 public abstract class Converter {
 
     private static final boolean NOT_NULL = false;
+    private static final boolean IS_NULL = true;
 
     // TODO - at the moment we haven't really formalised the idea of our internal format which we could do instead of
     //        defining the fields and the struct here.
 
     // Constants defining the fields used in the common output format.
     public static class ParsedDataFields {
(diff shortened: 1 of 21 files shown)


src/main/java/uk/gov/justice/digital/converter/Converter.java
```diff
@@ -6,32 +6,67 @@
...
+        public static final String TIMESTAMP = "_timestamp";
+        public static final String KEY = "_key";
+        public static final String DATA_TYPE = "_datatype";
+        public static final String OPERATION = "_operation";
+        public static final String SOURCE = "_source";
+        public static final String TABLE = "_table";
+        public static final String CONVERTER = "_converter";
     }
 
+    public enum Operation {
```
carwynmoj, line 40: Thanks for getting this in. This one has been bugging me for a while. I'll close the ticket for this one.

src/main/java/uk/gov/justice/digital/converter/Converter.java
```diff
@@ -6,32 +6,67 @@
...
+        Load("load"),
+        Insert("insert"),
```
carwynmoj, line 42: You don't actually need to define the name property since enums already have a `name()` method which you're using in `getOperation`.

src/main/java/uk/gov/justice/digital/service/DataStorageService.java
```diff
@@ -11,10 +11,18 @@
...
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
 
 @Singleton
 public class DataStorageService {
 
+    private static final String SOURCE = "source";
+    private static final String TARGET = "target";
+    private static final List<String> exclusions = Arrays.<String>asList();
```
carwynmoj, line 24: Doesn't look like this is used here so can be removed.

src/main/java/uk/gov/justice/digital/converter/Converter.java
dominic-messenger-moj: There are others coming such as copy-table, create-table etc which will need to be named differently. Will leave this in for now, in the expectation that the Operation and the name will not align exactly in future.

PR #75, opened 2023-05-30
DPR-478 integrate GlueSchemaClient into SourceReferenceService plus some refactoring

carwynmoj:
Summary of changes
* `SourceReferenceService` is now a singleton that is injected where used
* `SourceReferenceService` now uses the `GlueSchemaClient` to retrieve schemas, falling back to the original local resources where no schema is found
* `SourceReferenceService` now uses the `AvroToSparkSchemaConverter` to convert avro schema strings into a spark StructType - this uses the `spark-avro` library
* Refactored the `DataStorageService`
  * introducing the `SparkSessionProvider` so a spark session now does not need to be passed to those methods that need it
  * removed unnecessary null arg checks on some methods - reviewed where and how these methods are used and concluded it's very unlikely that we'd ever pass a null argument to one of these methods
  * tidied up the `DataStorageService` tests, removing duplication along with some test cases that were only exercising Mockito and also making use of mocks wherever possible reducing test runtime
  * also fixed `assertTrue(true)` where present replacing this with a relevant assert, e.g. verifying a mock was called where the method being tested is void
* other minor changes including
    * minor logging changes to clarify messages
    * revision to gradlew config following the gradle recommendations around the number of workers which actually does result in a reduction in overall test runtimes


src/main/java/uk/gov/justice/digital/service/DataStorageService.java
dominic-messenger-moj, line 1: Leave this as it was, for many Spark-related reasons.

dominic-messenger-moj (commented):
Please leave DataStorage service alone.
If it 'contains' a Spark Session it becomes impossible for it to manage writing on a distributed basis.
Second, data frames that are written have their own sessions.

src/main/java/uk/gov/justice/digital/service/DataStorageService.java
carwynmoj: No worries, will revert those changes.

carwynmoj: Ok, this has been done. Note that I've retained the logging changes and minor refactors. Shout if you have any concerns.

dominic-messenger-moj (approved):

PR #78, opened 2023-06-02
NOJIRA split DomainExecutorTest into three separate tests

carwynmoj:


carwynmoj:
No real wins here on circle ci. Does yield a reduction of ~20s locally but not really worth it. Back to the drawing board.

PR #84, opened 2023-06-07
Fix various data validation issues and improve logging

carwynmoj:
Summary of changes
* revisions to logging to improve visibility of validation errors
* introduced data filters (just for nomis for now) that tweak the raw data so it matches the parsed data returned by spark covering
  * timestamp formats 
  * date fields containing timestamp data (with and without time values)
* added proper handling of nullable avro fields 
  * our custom schema provides a `nullable` property which is now converted into the correct avro representation prior to conversion to a spark struct
* standardise on UTC to ensure consistent representation between raw data and that returned by spark

### DIFF
--- a/src/main/java/uk/gov/justice/digital/converter/avro/AvroToSparkSchemaConverter.java
+++ b/src/main/java/uk/gov/justice/digital/converter/avro/AvroToSparkSchemaConverter.java
@@ -1,13 +1,16 @@
 package uk.gov.justice.digital.converter.avro;
 
 import jakarta.inject.Singleton;
+import lombok.val;
 import org.apache.avro.Schema;
 import org.apache.spark.sql.avro.SchemaConverters;
 import org.apache.spark.sql.types.DataType;
 import org.apache.spark.sql.types.StructType;
 import uk.gov.justice.digital.converter.Converter;
 
+import java.util.List;
 import java.util.Optional;
+import java.util.stream.Collectors;
 
 /**
  * Avro Schema to Spark Schema (StructType) converter.
@@ -17,6 +20,10 @@ import java.util.Optional;
  * Avro types are converted to Spark types as per the documentation.
  * <p>
  * See <a href="https://spark.apache.org/docs/3.3.0/sql-data-sources-avro.html#supported-types-for-avro---spark-sql-conversion">Avro to Spark Schema Conversion Documentation</a>
+ * <p>
+ * Note - our custom `nullable` field property is converted to the correct avro representation by the schema converter
+ *        prior to conversion to the Spark Schema ensuring that nullable fields are represented correctly after
+ *        conversion.
  */
 @Singleton
 public class AvroToSparkSchemaConverter implements Converter<String, StructType> {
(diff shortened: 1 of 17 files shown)


src/test/java/uk/gov/justice/digital/config/BaseSparkTest.java
dominic-messenger-moj, line 1: Should we not also set the TimeZone here (just to keep the session in prod and test aligned) ?

dominic-messenger-moj (commented):
Is the 'NOMIS' fix global, or does it only apply to schemas that have 'NOMIS' as their source (ie should we change the schema from CNOMT3 to NOMIS as it should be compatible across NOMIS, CNOMPP, CNOMT2, CNOMT3)

src/main/java/uk/gov/justice/digital/converter/avro/AvroToSparkSchemaConverter.java
```diff
@@ -36,7 +43,36 @@ private Schema toAvroSchema(String avro) {
...
+    private List<Schema.Field> applyNullablePropertyToFields(Schema avro) {
+        return avro.getFields().stream()
+                .map(this::applyNullablePropertyToField)
+                .collect(Collectors.toList());
+    }
+
+    private Schema.Field applyNullablePropertyToField(Schema.Field avroField) {
+
+        val isNullable = Optional.ofNullable(avroField.getObjectProp("nullable"))
+                .map(nullable -> nullable.equals(true))
```
stalwartsolutions, line 70: Do you need this map?

stalwartsolutions (approved):

carwynmoj:
> Is the 'NOMIS' fix global, or does it only apply to schemas that have 'NOMIS' as their source (ie should we change the schema from CNOMT3 to NOMIS as it should be compatible across NOMIS, CNOMPP, CNOMT2, CNOMT3)

This is global at the moment. The work to generalise this out will be covered in a separate ticket that I will be creating shortly.

src/test/java/uk/gov/justice/digital/config/BaseSparkTest.java
carwynmoj: Since it had no impact on the tests I left it as is. However from a consistency standpoint it makes sense to include it should the tests start to cover this in future.

src/main/java/uk/gov/justice/digital/converter/avro/AvroToSparkSchemaConverter.java
carwynmoj: Sadly yes. `getObjectProp` returns `Object` so the map implicitly gives us the boolean we need.

src/main/java/uk/gov/justice/digital/job/filter/TimestampToDateFieldFilter.java
```diff
@@ -0,0 +1,58 @@
...
+
+        // The raw date value may be represented as an ISO date time with a zeroed time part eg. 2012-01-01T00:00:00Z
+        // Split the string on T so we get the date before the T, and the time after the T.
+        // If we get two values we assume we're parsing a date and time, otherwise treat the value as a date.
+        val dateParts = Arrays.asList(value.split("T"));
+
+        if (dateParts.size() == 2) {
+            val dateString = dateParts.get(0);
+            val timeString = dateParts.get(1);
+            if (timeString != null && (timeString.replaceAll("[^1-9]", "").compareTo("0") > 0))
```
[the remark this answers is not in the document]
carwynmoj: Not quite, the regex leaves any digit over zero e.g. 

```scala> time
val res16: String = 01:01:01Z

scala> time.replaceAll("[^1-9]", "")
val res17: String = 111
```
So "111" is greater than "0" which tells us that time time value is set.

The aim here was to relatively cheaply figure out if the time is set or not. So if any non-zero values are present we'll trigger the warning log. (Originally this was a rejection). 

Obviously if the timestamp is legitimately set to midnight at 0s 000ms then we evaluate as false but this is acceptable since in practice we're processing a number of records so we should get a range of values where the time part is used.

PR #87, opened 2023-06-08
DPR-521 Allow dates represented as datetimes to pass validation if time value is zeroed

carwynmoj:
Summary of changes
* revised the json comparision so we compare the data using Maps which contain the original and parsed data (where the avro schema has been applied)
* validator now reformats datetime strings in the original data where
  * the date string contains a `T`indicating that it might be an ISO 8601 date time string
  * the time part contains no digits except 0
  * otherwise the date value is left as is
 * the validator then continues as before comparing the original data with reformatted dates against the parsed data
 * this will then allow validation to pass where some dates were originally represented using date times
 * test updates to cover the new work


dominic-messenger-moj (approved):
This is fine for now. I'd like to maybe revisit this in the future when we ingest DPS data as the date problem may be different and we will need to parse the ISODateTime correctly rather than split on the T and check for 0.

PR #89, opened 2023-06-15
DPR-198: Process CDC events and archive to raw zone

koladeadewuyi-moj:
Changes included in this PR:
- Process CDC events [DPR-309](https://dsdmoj.atlassian.net/browse/DPR-309)
- Archive CDC events to the raw zone [DPR-310](https://dsdmoj.atlassian.net/browse/DPR-310)

[DPR-309]: https://dsdmoj.atlassian.net/browse/DPR-309?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[DPR-310]: https://dsdmoj.atlassian.net/browse/DPR-310?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/main/java/uk/gov/justice/digital/zone/RawZone.java
+++ b/src/main/java/uk/gov/justice/digital/zone/RawZone.java
@@ -16,11 +16,10 @@ import javax.inject.Singleton;
 
 import static org.apache.spark.sql.functions.*;
 import static uk.gov.justice.digital.common.ResourcePath.createValidatedPath;
-import static uk.gov.justice.digital.converter.dms.DMS_3_4_6.Operation.*;
 import static uk.gov.justice.digital.converter.dms.DMS_3_4_6.ParsedDataFields.*;
 
 @Singleton
-public class RawZone extends Zone {
+public class RawZone extends UnfilteredZone {
 
     public static final String PRIMARY_KEY_NAME = "id";
     private static final Logger logger = LoggerFactory.getLogger(RawZone.class);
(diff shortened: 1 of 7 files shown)


src/main/java/uk/gov/justice/digital/converter/dms/DMS_3_4_6.java
```diff
@@ -81,6 +82,8 @@ public static Optional<Operation> getOperation(String operation) {
         }
     }
 
+    public static DMS_3_4_6.Operation[] cdcOperations = { Insert, Update, Delete };
```
carwynmoj, line 85: I think it's better to have this as a property on the `Operation` enum. See my comment re. the set further down.

src/main/java/uk/gov/justice/digital/zone/RawZone.java
```diff
@@ -30,6 +34,11 @@ public class RawZone extends Zone {
     private final DataStorageService storage;
     private final SourceReferenceService sourceReferenceService;
 
+    private final Set<String> cdcOperationsSet = Arrays
```
carwynmoj, line 37: I think it'd be good to have this on the `Operation` enum. So we could have something like `getCdcOperations` that returns a set. 

The reasoning being that this way, we only need to define this information once, and the definition is located right next to the definiton of the values themselves.

src/main/java/uk/gov/justice/digital/zone/RawZone.java
```diff
@@ -40,58 +49,69 @@ public RawZone(JobArguments jobArguments,
...
         val startTime = System.currentTimeMillis();
 
-        String rowSource = table.getAs(SOURCE);
-        String rowTable = table.getAs(TABLE);
-        String rowOperation = table.getAs(OPERATION);
+        String rowSource = row.getAs(SOURCE);
+        String rowTable = row.getAs(TABLE);
 
-        logger.info("Processing batch with {} records for source: {} table: {}",
+        logger.info("Processing raw batch with {} records for source: {} table: {}",
```
carwynmoj, line 79: This is fine, but bear in mind that the class is also in the log output so we can see which zone it is already. I can show you some log examples if needed.

src/test/java/uk/gov/justice/digital/zone/RawZoneTest.java
```diff
@@ -2,85 +2,250 @@
...
 import static org.mockito.Mockito.doNothing;
 import static org.mockito.Mockito.when;
 import static uk.gov.justice.digital.common.ResourcePath.createValidatedPath;
+import static uk.gov.justice.digital.converter.dms.DMS_3_4_6.Operation.*;
+import static uk.gov.justice.digital.converter.dms.DMS_3_4_6.ParsedDataFields.OPERATION;
 import static uk.gov.justice.digital.zone.Fixtures.*;
 
 @ExtendWith(MockitoExtension.class)
-class RawZoneTest {
+class RawZoneTest extends BaseSparkTest {
```
carwynmoj, line 35: 👍 This makes sense and good to get some proper integration test coverage on this class.

carwynmoj (approved):
Looks good so far.

Added some observations. I reckon adding the cdc operations to the enum might be worth doing otherwise nothing else really needs changing.

dominic-messenger-moj (commented):
Raw shouldn't filter, and all rows should be written as they come in (so shouldn't be sorted either).

src/main/java/uk/gov/justice/digital/zone/RawZone.java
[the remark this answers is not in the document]
koladeadewuyi-moj: The changes have now been made to have a single process for the RawZone.

PR #98, opened 2023-06-20
DPR-311: Assert null columns are kept by structured zone

koladeadewuyi-moj:
This PR:
- Adds tests to check that null columns are kept by the structured zone in `StructuredZoneTest.java`


dominic-messenger-moj (approved):
LGTM

PR #102, opened 2023-06-22
DPR-198: Remove isCDC and refactor structured zones

koladeadewuyi-moj:
This PR revisits the use of the `isCDC` field:
- Removes the use of the boolean `isDCD` field by restructuring the structured zone(s) to use an inheritance approach
- The restructuring splits the structured zone into `StructuredZoneLoad` and `StructuredZoneCDC`
- Code common to the load and CDC structured zones is placed in abstract class `StructuredZone`
- Each structured zone implementation implements a `DeltaWriter` interface and overrides the `writeValidRecords(...)` method


src/main/java/uk/gov/justice/digital/zone/StructuredZone.java
```diff
@@ -181,95 +135,57 @@ private void handleInvalidRecords(SparkSession spark,
...
     }
 
-    private void appendDataAndUpdateManifestForTable(SparkSession spark,
-                                                     Dataset<Row> dataFrame,
-                                                     String tablePath,
-                                                     SourceReference.PrimaryKey primaryKey) throws DataStorageException {
-        logger.info("Appending {} records to deltalake table: {}", dataFrame.count(), tablePath);
-        storage.appendDistinct(tablePath, dataFrame.drop(OPERATION), primaryKey);
+    @Override
+    public void writeInvalidRecords(
```
carwynmoj, line 179: Ideally this shouldn't be part of the public api. Would probably mean an abstract class instead with protected methods though I think.

src/main/java/uk/gov/justice/digital/zone/StructuredZoneCDC.java
```diff
@@ -0,0 +1,116 @@
...
+    }
+
+    @Override
+    public Dataset<Row> process(SparkSession spark, Dataset<Row> dataFrame, Row table) throws DataStorageException {
+        val filteredRecords = dataFrame.filter(col(OPERATION).isin(cdcOperations));
+        return super.process(spark, filteredRecords, table);
+    }
+
+    @Override
+    public void writeValidRecords(
```
carwynmoj, line 48: As above, ideally this should not be public since it's only really `process` that's part of the public API here.

src/main/java/uk/gov/justice/digital/zone/DeltaWriter.java
```diff
@@ -0,0 +1,27 @@
+package uk.gov.justice.digital.zone;
+
+import org.apache.spark.sql.Dataset;
+import org.apache.spark.sql.Row;
+import org.apache.spark.sql.SparkSession;
+import uk.gov.justice.digital.domain.model.SourceReference;
+import uk.gov.justice.digital.exception.DataStorageException;
+import uk.gov.justice.digital.service.DataStorageService;
+
+public interface DeltaWriter {
```
kartthik-n, line 10: can we change this class to abstract class instead of interface?

src/main/java/uk/gov/justice/digital/zone/StructuredZoneCDC.java
```diff
@@ -0,0 +1,116 @@
...
+
+    @NotNull
+    private Consumer<Row> processRow(
+            SparkSession spark,
+            DataStorageService storage,
+            String tablePath,
+            SourceReference.PrimaryKey primaryKey
+    ) {
+        return row -> {
+            val optionalOperation = getOperation(row.getAs(OPERATION));
```
carwynmoj, line 70: Could do something like 

```
getOperation(...)
    .map( operation -> {
        try {
            writeRow(...)
        }
        catch (...) {
        }
   })
   .orElse( logger.error )
```

src/main/java/uk/gov/justice/digital/zone/StructuredZoneCDC.java
```diff
@@ -0,0 +1,116 @@
...
+            if (optionalOperation.isPresent()) {
+                val operation = optionalOperation.get();
+                try {
+                    writeRow(spark, storage, tablePath, primaryKey, operation, row);
+                } catch (DataStorageException ex) {
+                    logger.warn("Failed to {}: \n{}\nto {}", operation.getName(), row.json(), tablePath);
```
carwynmoj, line 76: Should this just be a warning or would error be better?

Either way, does it may sense to continue or should we abort on this exception?

src/main/java/uk/gov/justice/digital/zone/StructuredZoneLoad.java
```diff
@@ -0,0 +1,67 @@
...
+    }
+
+    @Override
+    public Dataset<Row> process(SparkSession spark, Dataset<Row> dataFrame, Row table) throws DataStorageException {
+        val filteredRecords = dataFrame.filter(col(OPERATION).equalTo(Load.getName()));
+        return super.process(spark, filteredRecords, table);
+    }
+
+    @Override
+    public void writeValidRecords(
```
carwynmoj, line 41: As previous comments. If these don't need to be public we should make them protected.

src/main/java/uk/gov/justice/digital/zone/StructuredZoneLoad.java
```diff
@@ -0,0 +1,67 @@
...
+            Dataset<Row> validRecords
+    ) throws DataStorageException {
+        logger.info("Appending {} records to deltalake table: {}", validRecords.count(), tablePath);
+        storage.appendDistinct(tablePath, validRecords.drop(OPERATION), primaryKey);
+        logger.info("Append completed successfully");
+        storage.updateDeltaManifestForTable(spark, tablePath);
+    }
+
+    @Override
+    public void writeInvalidRecords(
```
carwynmoj, line 55: As above.

src/main/java/uk/gov/justice/digital/zone/StructuredZone.java
```diff
@@ -78,96 +68,60 @@ public Dataset<Row> process(
...
-                .filter(col(VALID).equalTo(true))
-                .select(PARSED_DATA + ".*", OPERATION);
+        val missingSchemaRecords = dataFrame
+                .select(col(DATA), col(METADATA))
+                .withColumn(ERROR, lit(String.format("Schema does not exist for %s/%s", source, table)));
 
-        val validRecordsCount = validRecords.count();
+        writeInvalidRecords(spark, storage, createValidatedPath(violationsPath, source, table), missingSchemaRecords);
 
-        if (validRecordsCount > 0) {
```
kartthik-n, line 154: don't call the spark action here. instead check validRecords is empty or not.

validRecords.isEmpty()

carwynmoj (approved):
LGTM 👍 

Just some minor comments
* can we just have the process method public on the concrete Zone classes
* you could use more functional syntax when handling the option

src/main/java/uk/gov/justice/digital/zone/StructuredZone.java
```diff
@@ -181,95 +135,57 @@ private void handleInvalidRecords(SparkSession spark,
 
         if (invalidRecordsCount > 0) {
```
kartthik-n, line 136: same here 

invalidRecords.isEmpty()

src/main/java/uk/gov/justice/digital/zone/StructuredZone.java
```diff
@@ -181,95 +135,57 @@ private void handleInvalidRecords(SparkSession spark,
...
+            String destinationPath,
+            SourceReference.PrimaryKey primaryKey
+    ) throws DataStorageException {
+        val validRecords = dataFrame.filter(col(VALID).equalTo(true)).select(PARSED_DATA + ".*", OPERATION);
+        val validRecordsCount = validRecords.count();
 
-        val missingSchemaRecords = dataFrame
-                .select(col(DATA), col(METADATA))
-                .withColumn(ERROR, lit(String.format("Schema does not exist for %s/%s", source, table)));
+        if (validRecordsCount > 0) {
```
kartthik-n, line 151: same here validRecords.isEmpty()

src/main/java/uk/gov/justice/digital/zone/StructuredZoneCDC.java
koladeadewuyi-moj: We'll revisit the log levels for the monitoring and alerting.
When a writing a CDC record fails, the streaming job should continue.

src/main/java/uk/gov/justice/digital/zone/StructuredZone.java
koladeadewuyi-moj: We use the count in the log messages at many points in the code.
I agree with you. For performance reasons it would be preferred to not force spark to perform an action unless necessary. I intend to revisit the points where this is done when we get to a performance optimisation stage.

src/main/java/uk/gov/justice/digital/zone/StructuredZone.java
koladeadewuyi-moj: Same as previous comment at https://github.com/ministryofjustice/digital-prison-reporting-jobs/pull/102#discussion_r1238363116

src/main/java/uk/gov/justice/digital/zone/StructuredZone.java
koladeadewuyi-moj: Same as previous comment at https://github.com/ministryofjustice/digital-prison-reporting-jobs/pull/102#discussion_r1238363116

src/main/java/uk/gov/justice/digital/zone/StructuredZone.java
carwynmoj: Yeah, this is my bad. Would be good to retain the counts using a more efficient method since we can derive throughput metrics from these values.

src/main/java/uk/gov/justice/digital/zone/StructuredZoneCDC.java
koladeadewuyi-moj: The functional approach doesn't compile. I'll leave it for now and revisit after the curated zone task is complete.
Thanks

carwynmoj: Ok, no worries.

PR #103, opened 2023-06-23
DPR-314: Apply CDC to curated zone

koladeadewuyi-moj:
This PR:
- Applies the CDC events to the curated zone


src/main/java/uk/gov/justice/digital/job/DataHubJob.java
```diff
@@ -13,10 +13,7 @@
 import uk.gov.justice.digital.converter.Converter;
```
kartthik-n, line 13: writeInvalidRecords() method should be moved to Zone class as abstract method and implement the method in All zone class where Zone class is getting extended. Zone extending DeltaWriter is better

src/main/java/uk/gov/justice/digital/zone/CuratedZone.java
```diff
@@ -28,9 +25,11 @@ public class CuratedZone implements Zone {
     private final SourceReferenceService sourceReferenceService;
 
     @Inject
-    public CuratedZone(JobArguments jobArguments,
-                       DataStorageService storage,
-                       SourceReferenceService sourceReferenceService) {
+    public CuratedZone(
```
kartthik-n, line 28: CuratedZoneCDC extends CuratedZone is something incorrect here. We should have one base class and extend from one base class rather than applying multi level inheritance

src/main/java/uk/gov/justice/digital/zone/CuratedZoneLoad.java
```diff
@@ -0,0 +1,52 @@
...
+import uk.gov.justice.digital.service.DataStorageService;
+import uk.gov.justice.digital.service.SourceReferenceService;
+
+import javax.inject.Inject;
+import javax.inject.Singleton;
+
+import static uk.gov.justice.digital.converter.dms.DMS_3_4_6.ParsedDataFields.OPERATION;
+
+@Singleton
+public class CuratedZoneLoad extends CuratedZone {
```
kartthik-n, line 20: revisit all Zone inheritance please. We should have one base class extending across all Zones

src/main/java/uk/gov/justice/digital/zone/CuratedZoneLoad.java
```diff
@@ -0,0 +1,52 @@
...
+    @Override
+    protected void writeValidRecords(
+            SparkSession spark,
+            DataStorageService storage,
+            String tablePath,
+            SourceReference.PrimaryKey primaryKey,
+            Dataset<Row> validRecords
+    ) throws DataStorageException {
+        logger.info("Appending {} records to deltalake table: {}", validRecords.count(), tablePath);
+        storage.appendDistinct(tablePath, validRecords.drop(OPERATION), primaryKey);
```
kartthik-n, line 47: exception block should not do any action here (calling appendDistinct). Rather take it to the calling place of this method

src/main/java/uk/gov/justice/digital/job/DataHubJob.java
dominic-messenger-moj: I disagree Karthik. 
Zone is an interface. 
There is little that each Zone shares. Raw just appends; Structured handles validation and upsert/merges to storage. Curated currently looks like Structured but will functionally move away.
DeltaWriter should be changed to ZoneWriter, to avoid confusion.

src/main/java/uk/gov/justice/digital/zone/CuratedZone.java
dominic-messenger-moj: Again disagree.

src/main/java/uk/gov/justice/digital/zone/CuratedZoneLoad.java
koladeadewuyi-moj: This not an exception block. It is a method body.

src/main/java/uk/gov/justice/digital/job/DataHubJob.java
koladeadewuyi-moj: Java does not support inheriting multiple classes so making `Zone` an abstract class will not compile.

Not every zone writes invalid records so putting `writeInvalidRecords()` in `Zone` would require making `Zone` an abstract class as you suggested. Unfortunately, this will break the compilation as mentioned earlier.

src/main/java/uk/gov/justice/digital/zone/CuratedZone.java
koladeadewuyi-moj: The inheritance is logically correct.
`CuratedZoneCDC` is a `CuratedZone` is a `Zone`.
likewise
`CuratedZoneLoad` is a `CuratedZone` is a `Zone`.

As shown in the class diagram below. There are three types of zones (`RawZone`, `StructuredZone` and `CuratedZone`).
The `StructuredZone` and `CuratedZone` have two subtypes based on whether they are for Load or CDC.

```mermaid
classDiagram
RawZone --|> Zone : implements
StructuredZoneLoad --|> StructuredZone : extends
StructuredZoneCDC --|> StructuredZone : extends
StructuredZone --|> Zone : implements
CuratedZoneLoad --|> CuratedZone : extends
CuratedZoneCDC --|> CuratedZone : extends
CuratedZone --|> Zone : implements
```

src/main/java/uk/gov/justice/digital/zone/CuratedZoneLoad.java
koladeadewuyi-moj: Same as my previous comment at https://github.com/ministryofjustice/digital-prison-reporting-jobs/pull/103#discussion_r1239830426

carwynmoj (commented):
Have some general comments as per our chat on slack summarised below
* consider composing the ZoneWriters into a CuratedZone instead of a two level inheritance hierarchy (Easier to reason about and more flexible since we can use writers in other jobs easily)
* remove new line characters from the log messages - this can make it harder to parse them when we start looking at processing logs later
* minor - consider renaming your fixtures package to test instead - this can then be used as a home for test related code (plus it's a convention I've followed across our projects)

src/main/java/uk/gov/justice/digital/writer/curated/CuratedZoneCdcWriter.java
```diff
@@ -0,0 +1,37 @@
...
+import org.apache.spark.sql.Dataset;
+import org.apache.spark.sql.Row;
+import org.apache.spark.sql.SparkSession;
+import uk.gov.justice.digital.domain.model.SourceReference;
+import uk.gov.justice.digital.service.DataStorageService;
+import uk.gov.justice.digital.writer.Writer;
+
+import javax.inject.Inject;
+
+public class CuratedZoneCdcWriter extends Writer {
```
carwynmoj, line 12: If you add a `@Singleton` annotation to your writer classes micronaut can take care of wiring them in where used. See my earlier comment.

src/main/java/uk/gov/justice/digital/zone/curated/CuratedZoneLoad.java
```diff
@@ -0,0 +1,45 @@
...
+
+    private CuratedZoneLoad(
+            JobArguments jobArguments,
+            SourceReferenceService sourceReference,
+            Writer writer
+    ) {
+        super(jobArguments, sourceReference, writer);
+    }
+
+    private static Writer createWriter(DataStorageService storage) {
```
carwynmoj, line 36: If the writer classes are all annotated with `@Singleton` micronaut will take care of injecting these deps in. (See later comment).

src/main/java/uk/gov/justice/digital/writer/Writer.java
```diff
@@ -7,84 +7,91 @@
...
+            SparkSession spark,
+            String destinationPath,
+            Dataset<Row> invalidRecords
+    ) throws DataStorageException;
 
-    @Inject
-    public StructuredZoneCDC(
-            JobArguments jobArguments,
+    // static common methods
+    protected static void appendDistinctRecords(
```
carwynmoj, line 40: Very minor - I'd have just implemented these as protected unless there's a reason I'm missing for these to be static?

carwynmoj (approved):
LGTM 👍 

Approved with some minor comments that can be deferred to later work.

src/main/java/uk/gov/justice/digital/writer/curated/CuratedZoneCdcWriter.java
koladeadewuyi-moj: I did consider using the @Singleton DI approach but wasn't sure what the performance impact would be given that the writer is invoked in a couple of places. Something definitely worth revisiting during the performance tuning.

src/main/java/uk/gov/justice/digital/writer/Writer.java
koladeadewuyi-moj: The static is for performance reasons.

PR #108, opened 2023-06-30
DPR-198: Log and skip schema not found records in curated zone

koladeadewuyi-moj:
This PR logs a warning message and skips records where the schema could not be found during the curated zone processing.

This is to be revisited at https://dsdmoj.atlassian.net/browse/DPR-588


carwynmoj (approved):
LGTM 👍

PR #165, opened 2023-08-03
DPR-262: Incremental domain refresh

koladeadewuyi-moj:


src/main/java/uk/gov/justice/digital/job/DataHubJob.java
```diff
@@ -90,18 +98,21 @@ private void batchProcessor(JavaRDD<byte[]> batch) {
...
-                    curatedZoneLoad.process(spark, structuredLoadDataFrame, table);
+                    val structuredLoadDataFrame = structuredZoneLoad.process(spark, dataFrameForTable, tableInfo);
+                    curatedZoneLoad.process(spark, structuredLoadDataFrame, tableInfo);
 
-                    val structuredIncrementalDataFrame = structuredZoneCDC.process(spark, dataFrameForTable, table);
-                    curatedZoneCDC.process(spark, structuredIncrementalDataFrame, table);
+                    val structuredIncrementalDataFrame = structuredZoneCDC.process(spark, dataFrameForTable, tableInfo);
+                    val curatedCdcDataFrame = curatedZoneCDC.process(spark, structuredIncrementalDataFrame, tableInfo);
+
+                    if (!curatedCdcDataFrame.isEmpty()) domainService
```
koladeadewuyi-moj, line 114: Here is the entry point of the incremental domain refresh. The step takes the curated CDC dataframe as an input.

src/main/java/uk/gov/justice/digital/job/udf/JsonValidator.java
```diff
@@ -83,21 +83,21 @@ public static UserDefinedFunction createAndRegister(
...
+                    (UDF2<String, String, String>) (String originalJson, String parsedJson) ->
                         jsonValidator.validate(originalJson, parsedJson, schema),
-                        DataTypes.BooleanType
+                        DataTypes.StringType
                 )
             );
     }
 
-    public boolean validate(
+    public String validate(
```
koladeadewuyi-moj, line 93: To enhance debugging of schema violations within the structured zone, an error message describing the validation result is used instead of a boolean. When the error message is empty then the record is considered valid.

dominic-messenger-moj: Does this cover of DPR-532 or is it only part of it ?

dominic-messenger-moj (approved):
LGTM. Can we check that DPR-532 is also covered off by this work ?

src/main/java/uk/gov/justice/digital/job/udf/JsonValidator.java
koladeadewuyi-moj: It's only part of it.
At the moment resulting error would only be a single error summary e.g
`Record does not match schema nomis/offenders: some more detail here`
DPR-532 will be more detailed.

PR #195, opened 2023-08-15
DPR-262: Load adjoining row before applying view transform

koladeadewuyi-moj:
This PR implements incremental refresh of the domains using the incoming CDC event to know which domain to execute.
The steps are:
- For a CDC event received, get the domain definitions which contain the same source as the event
- For each domain definition found, use its SQL viewText to determine how to query the adjoining sources. e.g. if a domain definition relies on the sources `source_A` and `source_B`, and one already has `source_A` from the CDC event, one can then determine how to query the `source_B` in order to end up with `dataFrame_B` which when joined with `dataFrame_A` allows the to creation of the domain record via executing the SQL viewText.
- Delete/update/insert the domain record into s3
- Update the manifests


dominic-messenger-moj (approved):
Great work

PR #199, opened 2023-08-16
DPR-622: Make log level configurable

koladeadewuyi-moj:
This PR makes the log level configurable without the need to make code changes.


src/main/java/uk/gov/justice/digital/job/DataHubJob.java
dominic-messenger-moj, line 1: Do we need to stop CDC if we are testing in prod ?

dominic-messenger-moj (approved):

PR #200, opened 2023-08-21
DPR-622: Fix bug in get tables in batch and improve logs

koladeadewuyi-moj:
This PR fixes the dedupe login in getTablesInBatch and also improves the logs so they are distinguishable per zone and operation type.

### DIFF
--- a/src/main/java/uk/gov/justice/digital/job/DataHubJob.java
+++ b/src/main/java/uk/gov/justice/digital/job/DataHubJob.java
@@ -153,7 +153,7 @@ public class DataHubJob implements Runnable {
     private List<Row> getTablesInBatch(Dataset<Row> dataFrame) {
         return dataFrame
                 .select(TABLE, SOURCE, OPERATION)
-                .distinct()
+                .dropDuplicates(TABLE, SOURCE)
                 .collectAsList();
     }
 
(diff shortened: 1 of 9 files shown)


src/main/java/uk/gov/justice/digital/job/DataHubJob.java
```diff
@@ -153,7 +153,7 @@ public void run() {
     private List<Row> getTablesInBatch(Dataset<Row> dataFrame) {
         return dataFrame
                 .select(TABLE, SOURCE, OPERATION)
-                .distinct()
+                .dropDuplicates(TABLE, SOURCE)
```
koladeadewuyi-moj, line 156: Because the operation was factored into the uniqueness this was returning a list with duplicate tables one for CDC and another for Load.

dominic-messenger-moj (approved):
LGTM.

src/main/java/uk/gov/justice/digital/job/DataHubJob.java
tom-ogle-moj: I wonder if we should add some kind of test to prevent regressions

koladeadewuyi-moj: Good Idea. Will do so.

PR #211, opened 2023-09-01
DPR-674: Fix bug in JsonValidator

koladeadewuyi-moj:
This fixes the bug in the JsonValidator where nullable columns missing in the raw data were treated as mandatory.

After the fix, 
```
{
    "mandatoryField": "foo"
}
```
is now treated as valid and equal to
```
{
    "mandatoryField": "foo",
    "nullableField": null 
}
```


dominic-messenger-moj (approved):
Looks great.

PR #215, opened 2023-09-03
DPR-677: Set stopGracefullyOnShutdown to true

koladeadewuyi-moj:
This PR allows a graceful shutdown of the spark streaming context when a Glue job termination is invoked.
This will prevent data loss and the Kinesis DynamoDB checkpoint from getting corrupted due to an abrupt shutdown.


dominic-messenger-moj (approved):
LGTM

PR #221, opened 2023-09-07
DPR-674: Use map transformation for validation

koladeadewuyi-moj:
This uses a map function instead of a UDF for the Json validation.

To avoid serialization errors, a static class JsonValidatorStatic is used.


tom-ogle-moj (approved):
Looks good to me

koladeadewuyi-moj:
> Since the JsonValidator class in the uk.gov.justice.digital.job.ud package is not used now, can we deleted it?

Good catch. The class must have resurfaced after the rebase on main branch.
I've deleted it now. Thanks

PR #227, opened 2023-09-19
DPR2-123: Support viewText with three way join

koladeadewuyi-moj:
This PR adds support for domains with joins on multiple tables.
For example, a domain containing three tables `tableA`, `tableB` and `tableC`
Where:
1.  `tableA` has a join with `tableB` but not with `tableC`
2. `tableC` has a join with `tableB` but not with `tableA`

Since there is no direct relationship between `tableA` and `tableC` we have to go through `tableB`.

The processing steps will be:
- When a CDC record for `tableA` arrives
- Then the dataframe record for `tableB` is loaded using the CDC record for `tableA`
- And the dataframe record for `tableC` is loaded using previously loaded record for `tableB`
- And the domain is refreshed

### DIFF
--- a/src/main/java/uk/gov/justice/digital/common/ColumnMapping.java
+++ b/src/main/java/uk/gov/justice/digital/common/ColumnMapping.java
@@ -1,0 +1,53 @@
+package uk.gov.justice.digital.common;
+
+import java.util.Objects;
+
+public class ColumnMapping {
+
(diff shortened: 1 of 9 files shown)


dominic-messenger-moj (approved):
Approved. Will need a walkthrough of ACs to ensure regressions are taken into account

PR #237, opened 2023-09-26
DPR2-110: Implement Glue Kinesis structured streaming data source

tom-ogle-moj:
- Add dependency on AWS Glue ETL library for development.
- Use the Glue Kinesis structured streaming source to avoid data loss and use GlueContext and Glue Job APIs to start the streaming job.
- Move batch processing to its own class and replace KinesisReader with KinesisDataProvider.
- Change DMS_3_4_7 converter to work with dataset in the format as provided by the Kinesis source and to output raw in the semi-structured format as provided by the Kinesis source rather than as a raw String.
- Change job arguments to be appropriate for Glue source rather than DStream source, e.g. Kinesis ARN vs stream name
- Add ability to optionally set debug level logging via glue arg. Previously this caused it to default to WARN.
- Add ability to set start position in Kinesis stream when no checkpoints exist (handy for speeding up testing).
- Create Glue and Spark context/session/etc. only once with the same SparkConf and explicitly inject it into batch processing rather than via DI framework.
- Key column now uses md5 of raw to_json rather than md5 of raw since raw column is no longer a string
- Update com.github.johnrengelman.shadow (the old version somehow got corrupted in the CircleCI cache and caused builds to fail, as we have seen previously).


koladeadewuyi-moj (approved):
Looks good to me. 👍

PR #241, opened 2023-10-05
DPR2-128: Skip failed maintenance and error only at the end.

tom-ogle-moj:


src/main/java/uk/gov/justice/digital/service/MaintenanceService.java
```diff
@@ -24,26 +27,52 @@ public MaintenanceService(DataStorageService storageService) {
...
      */
-    public void compactDeltaTables(SparkSession spark, String rootPath) throws DataStorageException {
+    public void compactDeltaTables(SparkSession spark, String rootPath) throws DataStorageException, MaintenanceOperationFailedException {
         logger.info("Beginning delta table compaction for tables under root path: {}", rootPath);
         List<String> deltaTablePaths = storageService.listDeltaTablePaths(spark, rootPath);
-        logger.info("Found delta tables at: {}", String.join(", ", (deltaTablePaths)));
-        for (String deltaTablePath : deltaTablePaths) {
-            storageService.compactDeltaTable(spark, deltaTablePath);
-        }
+        logger.info("Found delta tables at: {}", java.lang.String.join(", ", (deltaTablePaths)));
```
gavriil-g-moj, line 33: This was already there so not a new change but does this cause too much logging or are the delta table paths not so many to flood the logs?

gavriil-g-moj (approved):
LGTM

koladeadewuyi-moj (approved):

src/main/java/uk/gov/justice/digital/service/MaintenanceService.java
tom-ogle-moj: Good point. It is 6 right now but could be a lot more in the future, in which case the string join would be nasty. I'm not sure why intellij didn't fix the java.lang.String import either. 

How about I change it to log the number of tables found instead for both compact and vacuum?

gavriil-g-moj: Changing it to log the number of tables sounds like a good idea. 
If the actual paths are needed for debugging purposes maybe this log can be changed to debug level.

tom-ogle-moj: Have done so now

gavriil-g-moj (approved):

PR #242, opened 2023-10-06
DPR2-52: Allow maintenance jobs to recurse in to S3 "directories"

tom-ogle-moj:
- Maintenance jobs can now recurse into S3 "directories" below the root to look for nested delta lake tables
- Recursing 2 levels deep works for domain (2 levels of nesting with a table in a domain, e.g. the establishment table in the establishment domain is s3://dpr-domain-development/establishment/establishment/ where s3://dpr-domain-development/ is the root)
- Still works for raw, structured, curated

### DIFF
--- a/src/it/java/uk/gov/justice/digital/test/DeltaTablesTestBase.java
+++ b/src/it/java/uk/gov/justice/digital/test/DeltaTablesTestBase.java
@@ -17,27 +17,36 @@ import static uk.gov.justice.digital.test.SparkTestHelpers.*;
  */
 public class DeltaTablesTestBase extends BaseSparkTest {
     @TempDir
-    protected static Path rootPath;
-    protected static Path offendersTablePath;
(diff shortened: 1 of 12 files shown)


src/it/java/uk/gov/justice/digital/test/DeltaTablesTestBase.java
```diff
@@ -17,27 +17,36 @@
...
-    protected static Path offendersTablePath;
-    protected static Path offenderBookingsTablePath;
+    protected Path rootPath;
+    protected Path offendersTablePath;
+    protected Path offenderBookingsTablePath;
+    protected Path agencyLocationsTablePath;
+    protected Path internalLocationsTablePath;
 
 
-    protected static void setupDeltaTablesFixture() {
```
tom-ogle-moj, line 25: Need to write these per test now rather than per test class

koladeadewuyi-moj (approved):

PR #245, opened 2023-10-12
DPR2-151 write violations on retry failure

tom-ogle-moj:
- DataStorageService now throws DataStorageRetriesExhaustedException when retries exceeded instead of propagating original exception
- Introduce ViolationService to handle violations
- Catch DataStorageRetriesExhaustedException and write violations in each zone (allowing recording the zone within the violations table error message)
- Ensure DataStorageRetriesExhaustedException is propagated upwards - this also involved switching some lambda foreach, etc. out for for loops to allow checked exceptions to propagate

### DIFF
--- a/src/main/java/uk/gov/justice/digital/exception/DataStorageRetriesExhaustedException.java
+++ b/src/main/java/uk/gov/justice/digital/exception/DataStorageRetriesExhaustedException.java
@@ -6,6 +6,7 @@ package uk.gov.justice.digital.exception;
 public class DataStorageRetriesExhaustedException extends DataStorageException {
 
     private static final long serialVersionUID = -2122315875377665022L;
+
     public DataStorageRetriesExhaustedException(Throwable cause) {
         super(cause);
     }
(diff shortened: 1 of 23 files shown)


src/main/java/uk/gov/justice/digital/service/ViolationService.java
dominic-messenger-moj, line 1: Good stuff. This is what was needed all along!

src/main/java/uk/gov/justice/digital/exception/DataStorageRetriesExhaustedException.java
```diff
@@ -0,0 +1,10 @@
+package uk.gov.justice.digital.exception;
+
+/**
+ * Thrown when the DataStorageService exceeds the configured number of retries/attempts.
+ */
+public class DataStorageRetriesExhaustedException extends DataStorageException {
```
dominic-messenger-moj, line 6: Needs a serialization serialVersionUID as it is a serializable class

dominic-messenger-moj (approved):
Minor comment on DataStorageRetriesExhaustedException but still approve.

PR #250, opened 2023-10-31
DPR2-184: Write inserts, updates and deletes in batches

koladeadewuyi-moj:
This PR
- Selects the latest records in a CDC batch when there is more than one for the same primary key
- Writes the entire batch using a single merge operation
- The records are ordered by ascending timestamp so the most recent record would be applied last

### DIFF
--- a/src/main/java/uk/gov/justice/digital/writer/Writer.java
+++ b/src/main/java/uk/gov/justice/digital/writer/Writer.java
@@ -1,9 +1,6 @@
 package uk.gov.justice.digital.writer;
 
-import lombok.val;
-import org.apache.spark.sql.Dataset;
-import org.apache.spark.sql.Row;
-import org.apache.spark.sql.SparkSession;
+import org.apache.spark.sql.*;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import uk.gov.justice.digital.domain.model.SourceReference;
@@ -11,8 +8,8 @@ import uk.gov.justice.digital.exception.DataStorageException;
 import uk.gov.justice.digital.exception.DataStorageRetriesExhaustedException;
 import uk.gov.justice.digital.service.DataStorageService;
 
-import static org.apache.spark.sql.functions.col;
-import static uk.gov.justice.digital.converter.dms.DMS_3_4_7.Operation.*;
+import java.util.Arrays;
+
 import static uk.gov.justice.digital.converter.dms.DMS_3_4_7.ParsedDataFields.OPERATION;
 import static uk.gov.justice.digital.converter.dms.DMS_3_4_7.ParsedDataFields.TIMESTAMP;
 
@@ -20,7 +17,7 @@ public abstract class Writer {
 
     private static final Logger logger = LoggerFactory.getLogger(Writer.class);
 
-    public abstract void writeValidRecords(
+    public abstract Dataset<Row> writeValidRecords(
             SparkSession spark,
             String destinationPath,
             SourceReference.PrimaryKey primaryKey,
@@ -58,13 +55,7 @@ public abstract class Writer {
     ) throws DataStorageRetriesExhaustedException {
         logger.info("Applying {} CDC records to deltalake table: {}", validRecords.count(), destinationPath);
 
-        val inserts = validRecords.filter(col(OPERATION).equalTo(Insert.getName())).orderBy(TIMESTAMP);
-        val updates = validRecords.filter(col(OPERATION).equalTo(Update.getName())).orderBy(TIMESTAMP);
-        val deletes = validRecords.filter(col(OPERATION).equalTo(Delete.getName())).orderBy(TIMESTAMP);
-
-        storage.upsertRecords(spark, destinationPath, inserts.drop(OPERATION, TIMESTAMP), primaryKey);
-        storage.updateRecords(spark, destinationPath, updates.drop(OPERATION, TIMESTAMP), primaryKey);
-        storage.deleteRecords(spark, destinationPath, deletes.drop(OPERATION, TIMESTAMP), primaryKey);
+        storage.mergeRecords(spark, destinationPath, validRecords, primaryKey, Arrays.asList(OPERATION, TIMESTAMP));
 
         logger.info("CDC records successfully applied to table: {}", destinationPath);
         storage.updateDeltaManifestForTable(spark, destinationPath);
--- a/src/main/java/uk/gov/justice/digital/writer/Writer.java
+++ b/src/main/java/uk/gov/justice/digital/writer/Writer.java
@@ -4,20 +4,17 @@ import lombok.val;
 import org.apache.spark.sql.Dataset;
 import org.apache.spark.sql.Row;
 import org.apache.spark.sql.SparkSession;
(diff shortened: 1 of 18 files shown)


src/main/java/uk/gov/justice/digital/writer/Writer.java
```diff
@@ -61,70 +58,16 @@ protected static void writeCdcRecords(
...
 
-        for (Row row: validRecords.collectAsList()) {
-            processRow(spark, storage, destinationPath, primaryKey, row);
-        }
+        val inserts = validRecords.filter(col(OPERATION).equalTo(Insert.getName())).orderBy(TIMESTAMP);
+        val updates = validRecords.filter(col(OPERATION).equalTo(Update.getName())).orderBy(TIMESTAMP);
+        val deletes = validRecords.filter(col(OPERATION).equalTo(Delete.getName())).orderBy(TIMESTAMP);
+
+        storage.upsertRecords(spark, destinationPath, inserts.drop(OPERATION, TIMESTAMP), primaryKey);
+        storage.updateRecords(spark, destinationPath, updates.drop(OPERATION, TIMESTAMP), primaryKey);
```
tom-ogle-moj, line 66: What happens when there are multiple updates for the same PK? Does it apply them all in timestamp order?

src/main/java/uk/gov/justice/digital/writer/Writer.java
```diff
@@ -61,70 +58,16 @@ protected static void writeCdcRecords(
     ) throws DataStorageRetriesExhaustedException {
         logger.info("Applying {} CDC records to deltalake table: {}", validRecords.count(), destinationPath);
 
-        for (Row row: validRecords.collectAsList()) {
-            processRow(spark, storage, destinationPath, primaryKey, row);
-        }
+        val inserts = validRecords.filter(col(OPERATION).equalTo(Insert.getName())).orderBy(TIMESTAMP);
+        val updates = validRecords.filter(col(OPERATION).equalTo(Update.getName())).orderBy(TIMESTAMP);
+        val deletes = validRecords.filter(col(OPERATION).equalTo(Delete.getName())).orderBy(TIMESTAMP);
+
```
tom-ogle-moj, line 64: Are there any corner cases we need to worry about? This one is a bit of a stretch but, say you had a table with a PK that is a meaningful business key rather than something like an auto increment or uuid. Someone deletes it and recreates it with the same PK rather than updates it for some reason. I think you could then end up only deleting the row and not recreating it if it arrived in one batch?

src/main/java/uk/gov/justice/digital/writer/Writer.java
koladeadewuyi-moj: Good question. I think Delta has a feature for this. Taking a look.

tom-ogle-moj: If not and if every update contains the full up to date record we could probably drop all but the latest row for a given PK using a window function or something.

src/main/java/uk/gov/justice/digital/writer/Writer.java
dominic-messenger-moj: I agree. They need to be kept together.

src/main/java/uk/gov/justice/digital/writer/Writer.java
koladeadewuyi-moj: This is now implemented. Before writing, the latest records are selected when there are more than one record for a primary key.

src/main/java/uk/gov/justice/digital/writer/Writer.java
koladeadewuyi-moj: The entire CDC batch is now written in the same merge operation.

src/test/java/uk/gov/justice/digital/service/DataStorageServiceTest.java
```diff
@@ -288,21 +288,6 @@ public void shouldRetryAppendDistinctAndSucceedEventually() throws DataStorageEx
         verify(mockDeltaMergeBuilder, times(2)).execute();
     }
 
-    @Test
-    public void shouldRetryUpsertRecordsAndSucceedEventually() throws DataStorageRetriesExhaustedException {
```
koladeadewuyi-moj, line 292: Both upsert and delete now use the same merge operation leaving us with two duplicate tests which have now been removed.

src/test/java/uk/gov/justice/digital/service/DataStorageServiceTest.java
```diff
@@ -425,24 +412,6 @@ public void shouldRetryUpdateRecordsAndFailEventually() {
         verify(mockDeltaMergeBuilder, times(retryAttempts)).execute();
     }
 
-    @Test
-    public void shouldRetryDeleteRecordsAndFailEventually() {
```
koladeadewuyi-moj, line 429: This became a duplicate test and was removed.

src/test/java/uk/gov/justice/digital/service/DataStorageServiceTest.java
```diff
@@ -557,13 +528,6 @@ private void stubUpdateRecords() {
         when(mockDeltaMergeMatchedActionBuilder.updateAll()).thenReturn(mockDeltaMergeBuilder);
     }
 
-    private void stubDeleteRecords() {
```
koladeadewuyi-moj, line 560: This is now superseded by `stubMergeRecoords()`

dominic-messenger-moj (approved):
LGTM
My only question is - are _operation and _timestamp the only 2 fields we want to exclude

tom-ogle-moj (approved):

src/main/java/uk/gov/justice/digital/writer/structured/StructuredZoneCdcWriter.java
```diff
@@ -49,4 +59,31 @@ public void writeInvalidRecords(
...
+
+        val mostRecentColumnExpressionsAsScalaSeq = JavaConverters
+                .asScalaIteratorConverter(mostRecentColumnExpressions.iterator())
+                .asScala()
+                .toSeq();
+
+        if (mostRecentColumnExpressionsAsScalaSeq.nonEmpty()) {
+            return validRecords
+                    .orderBy(TIMESTAMP)
+                    .groupBy(primaryKeys)
```
tom-ogle-moj, line 80: @koladeadewuyi-moj I have seen some conflicting accounts on whether a DataFrame orderBy followed by a groupBy always preserves the order within the groups and under what circumstances the order is preserved. For example, I've seen claims that if the data was not already partitioned by the grouping key then the order may be jumbled up again during the shuffle to produce the groups and therefore you should also include the grouping keys in the orderby. I think I've also seen claims that even that isn't good enough to guarantee the order is preserved in the groups. I think something also said that if the data spills to disk during the grouping it could break the ordering.

https://medium.com/@davisjustin42/group-by-does-not-maintain-order-in-pyspark-use-a-window-function-instead-9178e70497a9

https://stackoverflow.com/questions/39505599/spark-dataframe-does-groupby-after-orderby-maintain-that-order

Is there any likelihood of taking a value for a column here that isn't from the last row?

PR #255, opened 2023-11-20
DPR2-208 Streaming CDC Job

tom-ogle-moj:


src/main/java/uk/gov/justice/digital/client/s3/S3DataProvider.java
```diff
@@ -31,7 +31,7 @@ public S3DataProvider(JobArguments arguments) {
 
     public Dataset<Row> getSourceData(SparkSession sparkSession, String schemaName, String tableName) {
         String tablePath = tablePath(arguments.getRawS3Path(), schemaName, tableName);
-        String fileGlobPath = ensureEndsWithSlash(tablePath) + arguments.getBatchLoadFileGlobPattern();
+        String fileGlobPath = ensureEndsWithSlash(tablePath) + arguments.getCdcFileGlobPattern();
```
tom-ogle-moj, line 34: This was a mistake in the batch job PR - the whole class should have been included in this PR instead with the CDC file glob pattern.

src/main/java/uk/gov/justice/digital/job/DataHubCdcJob.java
```diff
@@ -0,0 +1,115 @@
...
+        if(!tablesToProcess.isEmpty()) {
+            tablesToProcess.forEach(tableDetails -> {
+                String inputSchemaName = tableDetails.getLeft();
+                String inputTableName = tableDetails.getRight();
+                TableStreamingQuery streamingQuery = tableStreamingQueryProvider.provide(inputSchemaName, inputTableName);
+                streamingQuery.runQuery(spark);
+                streamingQueries.add(streamingQuery);
+            });
+        } else {
+            logger.warn("No tables to process");
```
tom-ogle-moj, line 101: I was unsure of whether to throw and crash out here instead but I figured you might conceivably legitimately deploy a streaming app for some set of tables and only later add the configs for the tables and restart it.

src/it/java/uk/gov/justice/digital/job/cdc/TableStreamingQueryIT.java
```diff
@@ -0,0 +1,266 @@
...
+        streamingQuery = underTest.runQuery(spark);
+    }
+
+    private void whenTheNextBatchIsProcessed() {
+        streamingQuery.processAllAvailable();
+    }
+
+    private void whenInsertOccursForPK(int primaryKey, String data, String timestamp) {
+        List<Row> input = Collections.singletonList(
+                RowFactory.create(Integer.toString(primaryKey), timestamp, "I", data)
```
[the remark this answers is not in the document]
tom-ogle-moj: Yeah, good point. I will do so here and search through for other places it makes sense.

src/main/java/uk/gov/justice/digital/job/DataHubCdcJob.java
koladeadewuyi-moj: I wonder if it may mask a misconfiguration. I guess we could verify from the logs if a table is being processed.

koladeadewuyi-moj (approved):

src/it/java/uk/gov/justice/digital/job/cdc/TableStreamingQueryIT.java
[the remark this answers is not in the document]
tom-ogle-moj: I've done this now

PR #257, opened 2023-11-24
DPR2-215: Use avro actual schemas rather than inferring.

tom-ogle-moj:
Also remove distinction between input source/table (e.g. oms_owner) and sourceReference source/table (e.g. nomis) since they are now unified by DMS.


src/main/java/uk/gov/justice/digital/job/cdc/TableStreamingQuery.java
```diff
@@ -27,8 +27,6 @@ public class TableStreamingQuery {
     private final JobArguments arguments;
     private final S3DataProvider s3DataProvider;
     private final CdcBatchProcessor batchProcessor;
-    private final String inputSchemaName;
```
tom-ogle-moj, line 30: Now that we don't use both nomis and oms_owner in different places the distinction between these Strings and the source and table inside source reference is moot - so I have removed it.

koladeadewuyi-moj (approved):

PR #258, opened 2023-11-27
DPR-209: Add hive table creation job

koladeadewuyi-moj:
This PR:
- Adds a job to create the hive tables for every schema in the schema registry
- Removes the domain refresh logic

Note: I've had to revert the configuration which makes structured streaming archive files after processing for because it can only archive within the same bucket.
- `s3://dpr-raw-zone-development/nomis/agency_internal_locations/*-*.parquet` -> `s3://dpr-raw-archive-development/nomis/agency_internal_locations/*-*.parquet` failed
- `s3://dpr-raw-zone-development/nomis/agency_internal_locations/*-*.parquet` -> `s3://dpr-raw-archive-development/archive/nomis/agency_internal_locations/*-*.parquet` failed
- `s3://dpr-raw-zone-development/nomis/agency_internal_locations/*-*.parquet` -> `s3://dpr-raw-zone-development/archive/nomis/agency_internal_locations/*-*.parquet` worked

Leaving out the wildcard pattern from the destination directory didn't work either.

It should be noted that structured streaming only moves files when newer ones arrive and since it uses the spark threads may also have performance implications on the streaming job. Also, we have a lambda function which periodically moves the files which may be sufficient.


src/main/java/uk/gov/justice/digital/client/glue/GlueHiveTableClient.java
dominic-messenger-moj, line 1: Do we need to do anything about date and timestamp columns ?

src/main/java/uk/gov/justice/digital/job/HiveTableCreationJob.java
dominic-messenger-moj, line 1: Interested to know how this will be run...
1) As part of the load for a specific schema group; or
2) For all schemas.

dominic-messenger-moj (approved):
Quite extensive changes. From a code perspective all good.

src/main/java/uk/gov/justice/digital/client/glue/GlueHiveTableClient.java
koladeadewuyi-moj: I don't believe so. The getColumnsAndModifyTypes method does not modify the timestamp columns.

src/main/java/uk/gov/justice/digital/job/HiveTableCreationJob.java
koladeadewuyi-moj: Currently for all available schemas.

PR #269, opened 2023-12-08
DPR2-210: Change violations schema to support writing datatype changes to violations

tom-ogle-moj:
- The violations tables are currently using the schemas of the input tables. 
- This is incompatible with writing data that doesn't match the schema (in certain incompatible ways, such as data type changes). If the data doesn't match the schema then you can't write it to a table that requires that schema!
- This work changes the violations table schemas to allow violations to be written when the schema has changed in an incompatible way since the last write to violations, such as incompatible data type change
- To enable this it changes the schema of violations from that of the input table to two columns: a string containing the error and a string containing the original input converted to json.
- Added tests to show violations with different input schemas can be written to the same violations table
- Also added a test to show that violations data with different input schemas can subsequently be read back in using spark with the data types widening to string if the schemas are incompatible.


dominic-messenger-moj (approved):
LGTM

PR #272, opened 2023-12-12
DPR2-245: Read schemas from s3

koladeadewuyi-moj:
This PR reads the schemas from an S3 bucket instead of Glue schema registry.


domains/domains.sql
tom-ogle-moj, line 1: Are you purposely deleting this file or is it an accident from a rebase or something?

tom-ogle-moj (approved):

PR #275, opened 2023-12-13
DPR2-210: Solve Scala java integration issue

tom-ogle-moj:
- When Spark reads data it can throw a SparkException (a checked Exception type)
- It does not advertise that it throws this using the scala @throws annotation which means that the bytecode scala generates does not have the equivalent of `throws SparkException`
- This means that the java compiler does not believe the SparkException can be thrown.
- It seems that the `catch (SparkException e)` block is optimised away because Java thinks this block can never be reached.
- The previous code works when the main method is run against an environment using Intellij but does not catch the same exception for the same input data when run in Glue using a packaged jar.
- This change changes the catch to catch a broader exception type.
- It also hides away the scala - java integration nastiness inside the data provider

### DIFF
--- a/src/main/java/uk/gov/justice/digital/client/s3/S3DataProvider.java
+++ b/src/main/java/uk/gov/justice/digital/client/s3/S3DataProvider.java
@@ -3,7 +3,6 @@ package uk.gov.justice.digital.client.s3;
 import jakarta.inject.Inject;
 import jakarta.inject.Singleton;
 import lombok.val;
-import org.apache.spark.SparkException;
 import org.apache.spark.sql.AnalysisException;
 import org.apache.spark.sql.Dataset;
 import org.apache.spark.sql.Row;
@@ -14,7 +13,7 @@ import org.slf4j.LoggerFactory;
 import scala.collection.JavaConverters;
 import uk.gov.justice.digital.config.JobArguments;
 import uk.gov.justice.digital.domain.model.SourceReference;
-import uk.gov.justice.digital.exception.FailedMergingSchemas;
+import uk.gov.justice.digital.exception.DataProviderFailedMergingSchemasException;
 import uk.gov.justice.digital.exception.NoSchemaNoDataException;
 
 import java.util.List;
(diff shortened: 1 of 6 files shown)


src/main/java/uk/gov/justice/digital/client/s3/S3DataProvider.java
```diff
@@ -80,12 +80,23 @@ public Dataset<Row> getStreamingSourceDataWithSchemaInference(SparkSession spark
...
-                .parquet(scalaFilePaths);
+    public Dataset<Row> getBatchSourceData(SparkSession sparkSession, List<String> filePaths) throws DataProviderFailedMergingSchemasException {
+        try {
+            val scalaFilePaths = JavaConverters.asScalaIteratorConverter(filePaths.iterator()).asScala().toSeq();
+            return sparkSession
+                    .read()
+                    .option("mergeSchema", "true")
+                    .parquet(scalaFilePaths);
+        } catch (Exception e) {
+            // We can't catch the specific SparkException type we actually want here because scala erases the fact
```
tom-ogle-moj, line 91: I thought moving this scala vs java nastiness in here and throwing our own exception would keep the main part of the app a bit cleaner

src/main/java/uk/gov/justice/digital/job/DataHubBatchJob.java
```diff
@@ -144,14 +145,10 @@ private void processFilePaths(SparkSession sparkSession, String schema, String t
                 logger.warn("No source reference for table {}.{} - writing all data to violations", schema, table);
                 violationService.handleNoSchemaFoundS3(sparkSession, dataFrame, schema, table, STRUCTURED_LOAD);
             }
-        } catch (SparkException e) {
```
tom-ogle-moj, line 147: I've verified that SparkException is what is actually thrown both via stacktrace and debugging while running the app against the dev environment from intellij. But it just isn't caught when running the app using the jar. I'm fairly certain that the reason is Java optimising away code paths it thinks can never be reached.

stalwartsolutions (approved):

koladeadewuyi-moj (approved):

PR #276, opened 2023-12-14
DPR2-210: Handle nested incompatible schema exception.

tom-ogle-moj:
- In the dev environment Spark throws the SparkException indicating the incompatible schema and is handled as per the last PR.
- In the test environment, presumably due to the slightly different test data mixed in with my test case data, Spark is wrapping the exception in another one.
- This PR handles the wrapped case by checking the cause.


tom-ogle-moj:
I've run a locally generated build of this in test and it fixes the issue.

koladeadewuyi-moj (approved):

PR #277, opened 2023-12-18
DPR2-295 changes to movements domain to support hospital & courts

dominic-messenger-moj:


zahoor-moj (approved):
Domain includes additional columns for reason_code, escort and comment.

PR #284, opened 2024-01-15
DPR2-35: Delete duplicate test helpers

tom-ogle-moj:
Deletes ResourceLoader from integration tests that was duplicated from test. The other code duplicated between test and integration test has already been merged in to single classes in test.


gavriil-g-moj (approved):
LGTM

PR #294, opened 2024-01-25
DPR2-46: Update manifest after switching prisons data source

koladeadewuyi-moj:
This PR:
- Updates the delta manifest so the prisons table is still queryable after it has been switched to use the data located in the temp-reload bucket
- Renames HiveSchemaService to HiveTableService

### DIFF
--- a/src/main/java/uk/gov/justice/digital/service/HiveTableService.java
+++ b/src/main/java/uk/gov/justice/digital/service/HiveTableService.java
@@ -2,6 +2,7 @@ package uk.gov.justice.digital.service;
 
 import com.google.common.collect.ImmutableSet;
 import org.apache.commons.lang3.tuple.ImmutablePair;
+import org.apache.spark.sql.SparkSession;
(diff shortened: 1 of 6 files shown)


src/main/java/uk/gov/justice/digital/service/HiveTableService.java
```diff
@@ -101,6 +105,7 @@ public Set<ImmutablePair<String, String>> switchPrisonsTableDataSource(Immutable
                 String dataPath = createValidatedPath(targetS3Path, sourceName, tableName);
 
                 replaceSymlinkInputTables(jobArguments.getPrisonsDatabase(), hiveTableName, dataPath, schema);
+                storageService.updateDeltaManifestForTable(spark, dataPath);
```
koladeadewuyi-moj, line 108: This updates the delta manifest so the prisons table is still queryable after it has been switched to use the data located in the temp-reload bucket.

tom-ogle-moj (approved):

PR #297, opened 2024-02-26
DPR2-378: Special case to mark 8 bit ints as valid when the avro sche…

tom-ogle-moj:
…ma says (32 bit) int.

This is because avro does not allow smaller ints than 32 bit so we have to use 32 bit ints for 32 bit and smaller. Previously we special cased 16 bit ints but not 8 bit ints.


src/test/java/uk/gov/justice/digital/service/ValidationServiceSchemasMatchTest.java
```diff
@@ -10,9 +10,9 @@
 import static org.junit.jupiter.api.Assertions.assertTrue;
 import static uk.gov.justice.digital.service.ValidationService.schemasMatch;
 
-public class ValidationServiceSchemasMatchTest {
+class ValidationServiceSchemasMatchTest {
```
tom-ogle-moj, line 13: Removing public modifiers to satisfy Sonar quality issues

koladeadewuyi-moj (approved):

PR #301, opened 2024-04-26
DPR2-662: Expose cleanupSource, broadcastJoin settings

koladeadewuyi-moj:
This PR exposes the below configs so we can release to prod and adjust for the optimum values. The configs are given default values to maintain the existing behaviour:

To enable solving the errors caused by the broadcast timeouts in the movements CDC job the configs:
- `dpr.spark.broadcast.timeout.seconds` allows configuring the timeout to a higher value
- `dpr.disable.auto.broadcast.join.threshold` allows disabling of the broadcast join if necessary to see if it improves stability

To fix the interference from the S3 file archive job when moving files from the same location the CDC job is reading from the configs below were exposed:
- `dpr.enable.streaming.source.archiving` enables/disables moving of the processed files to a folder by the CDC job. Currently disabled to maintain existing behaviour
- `dpr.processed.raw.files.path` this is the folder to which the CDC job will move processed files to before they are finally archived by the S3 file archive job
- `dpr.file.source.prefix` this is exposed to allow the S3 file archive job source for the files from the processed files folder
- `dpr.file.transfer.destination.prefix` allows the S3 file archive job move files to a custom folder. In essence, the job removes the source prefix when writing to the raw-archive bucket


tom-ogle-moj (approved):
LGTM

PR #302, opened 2024-05-03
DPR2-20: Use a custom log4j.properties file in Glue

tom-ogle-moj:
Remove the setting of Spark log level in code now that we can set it with a custom log4j2.properties file.


build.gradle
```diff
@@ -112,6 +112,10 @@ configurations {
     integrationTestImplementation.extendsFrom implementation, testImplementation
 }
 
+configurations.all {
+    exclude group: 'org.slf4j', module: 'slf4j-reload4j'
```
tom-ogle-moj, line 116: Excludes a conflicting slf4j backend

src/main/java/uk/gov/justice/digital/provider/SparkSessionProvider.java
```diff
@@ -16,23 +16,17 @@ public class SparkSessionProvider {
     public GlueContext createGlueContext(String jobName, JobArguments arguments) {
         SparkConf sparkConf = new SparkConf().setAppName(jobName);
         SparkSessionProvider.configureSparkConf(sparkConf, arguments);
-        SparkContext spark = new SparkContext(sparkConf);
-        spark.setLogLevel(arguments.getLogLevel().name().toUpperCase());
```
tom-ogle-moj, line 20: This would stop the custom logging config file from taking effect

koladeadewuyi-moj (approved):

PR #313, opened 2024-06-25
DPR2-894 load tables to operational datastore

tom-ogle-moj:
This PR updates the DataHub batch job to:

1. Read the AWS Glue Connection details from the Glue API
2. Read the credentials in the secret referenced in the Glue Connection details from Secrets Manager
3. Apply some transformations that are required to write the data to Postgres
    - Remove null String characters
    - Normalise column names to lower-case
    - Remove the Op and _timestamp metadata columns

To do this it...:
- Adds a couple of JobArguments to configure the write to the operational datastore
- Adds OperationalDataStoreService to manage writing to the Operational Data Store
- Calls it from the BatchProcessor in the batch job with the result of the curated zone processing
- Adds some extra classes to support OperationalDataStoreService and to split the operational data store related functionality/responsibility
- Adds unit tests
- Adds integration tests

Also, if the write to the Operational DataStore is disabled (such as in environments beyond dev that don't currently have an ODS) then the app will still work due to a feature flag that defaults to disabled.


src/main/java/uk/gov/justice/digital/service/operationaldatastore/OperationalDataStoreTransformation.java
```diff
@@ -0,0 +1,52 @@
...
+    private static final Logger logger = LoggerFactory.getLogger(OperationalDataStoreTransformation.class);
+
+    Dataset<Row> transform(Dataset<Row> dataFrame) {
+        logger.debug("Setting up data transformations ready for Operational DataStore");
+        // We normalise columns to lower case to avoid having to quote every column due to Postgres lower casing everything in incoming queries
+        Dataset<Row> lowerCaseColsDf = normaliseColumnsToLowerCase(dataFrame);
+        // Handle 0x00 null String character which cannot be inserted in to a Postgres text column
+        Dataset<Row> withoutNullsDf = stripNullStrings(lowerCaseColsDf);
+        // We don't store these metadata columns in the destination table so we remove them
+        return withoutNullsDf.drop(OPERATION.toLowerCase(), TIMESTAMP.toLowerCase());
```
koladeadewuyi-moj, line 34: Do you intend to drop the `checkpoint_col` too?

tom-ogle-moj: @koladeadewuyi-moj Yes I will but I'm going to do it in a separate PR since this one is pretty big already

koladeadewuyi-moj (approved):

PR #316, opened 2024-07-02
DPR2-895: Add merge functionality to OperationalDataStoreService.

tom-ogle-moj:
- Update the CDC job so that on each batch it:
    - Writes the output of curated to a loading table (e.g. loading.offenders), overwriting any data if the table already exists.
    - Runs a SQL MERGE statement to merge the data in the loading table in to the destination table, which already exists since it was created by the batch job (e.g. from loading.offenders in to nomis.offenders).
    - Finally, it truncates the loading table.
- Add a JobArguments property to configure the loading schema name.
- Updated E2E and IT tests to use in-memory operational data store when testing CDC functionality
- Unit tests
- Removes some duplication from the E2E and IT tests in to SharedTestFunctions
- Moves the dropping of metadata columns out of the Operational DataStore transformer and in to the service since this functionality differs between batch and CDC

### DIFF
--- a/src/main/java/uk/gov/justice/digital/service/operationaldatastore/OperationalDataStoreTransformation.java
+++ b/src/main/java/uk/gov/justice/digital/service/operationaldatastore/OperationalDataStoreTransformation.java
@@ -13,9 +13,6 @@ import java.util.Arrays;
 
 import static org.apache.spark.sql.functions.col;
 import static org.apache.spark.sql.functions.regexp_replace;
-import static uk.gov.justice.digital.common.CommonDataFields.CHECKPOINT_COL;
-import static uk.gov.justice.digital.common.CommonDataFields.OPERATION;
-import static uk.gov.justice.digital.common.CommonDataFields.TIMESTAMP;
 
 /**
  * Transforms a DataFrame to the format we want to store in the Operational DataStore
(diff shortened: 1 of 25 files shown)


src/main/java/uk/gov/justice/digital/service/operationaldatastore/OperationalDataStoreTransformation.java
```diff
@@ -30,9 +27,7 @@ Dataset<Row> transform(Dataset<Row> dataFrame) {
         // We normalise columns to lower case to avoid having to quote every column due to Postgres lower casing everything in incoming queries
         Dataset<Row> lowerCaseColsDf = normaliseColumnsToLowerCase(dataFrame);
         // Handle 0x00 null String character which cannot be inserted in to a Postgres text column
-        Dataset<Row> withoutNullsDf = stripNullStrings(lowerCaseColsDf);
-        // We don't store these metadata columns in the destination table so we remove them
-        return withoutNullsDf.drop(OPERATION.toLowerCase(), TIMESTAMP.toLowerCase(), CHECKPOINT_COL.toLowerCase());
```
tom-ogle-moj, line 35: Most of the transformation is common between batch and CDC, but the columns are not (since CDC needs the op column in the temporary table for the merge)

koladeadewuyi-moj (approved):

PR #317, opened 2024-07-17
DPR2-986: Load the database driver explicitly each time we start a test database.

tom-ogle-moj:
- One of the integration tests fails in the release repo but not in the jobs repo.
- This seems to be because the release repo runs the tests in parallel using gradle and this causes some strangeness in class loading.
- I can reproduce the issue locally with `./gradlew --parallel --max-workers=4 clean build -x test`
- The changes in this PR fix the issue when run locally
- It loads the JDBC database driver explicitly for each test that uses the in memory database using Class.forName
- This shouldn't be necessary as it is the legacy way of ensuring the class is loaded but seems to fix the weird class loading issue


harichintala1 (approved):
LGTM

PR #319, opened 2024-07-18
DPR2-901 limit ods writes to configured tables

tom-ogle-moj:
- On app startup the set of tables to write to the operational datastore is read from the configuration.datahub_managed_tables table in the operational datastore, which specifies the tables the operational datastore should be managing
- The table name is configured using a new JobArgument `dpr.operational.data.store.tables.to.write.table.name`
- When the job calls the operational datastore service to overwrite or merge a table, it checks if the table should be written or not based on the set of tables read during startup


src/main/java/uk/gov/justice/digital/service/operationaldatastore/dataaccess/OperationalDataStoreRepositoryProvider.java
```diff
@@ -0,0 +1,20 @@
...
+@Singleton
+public class OperationalDataStoreRepositoryProvider {
+
+    private final JobArguments jobArguments;
+
+    public OperationalDataStoreRepositoryProvider(JobArguments jobArguments) {
+        this.jobArguments = jobArguments;
+    }
+
+    OperationalDataStoreRepository getOperationalDataStoreRepository(DataSource dataSource) {
```
tom-ogle-moj, line 17: The provider is there in order to share the connection pool and still be able to mock injected dependencies

koladeadewuyi-moj (approved):

PR #320, opened 2024-07-22
DPR2-714: Add batch job to create reload diffs

koladeadewuyi-moj:
This PR adds Glue job which creates diffs between the reloaded dataset in the raw bucket and the archived dataset. The diffs created are placed in the `s3://dpr-temp-reload-<env>/diffs/` folder:

- `toDelete`: These are the records not present in the reload data but available in the archive data
- `toInsert`: These are the records present in the reload data but not available in the archive data
- `toUpdate`: These are the records present in both the reload data and the archive but which have different column values (columns compared are the ones defined in the contracts)

Each diff type contains parquet data with records having an appropriate operation code (toDelete -> Delete, toInsert -> Insert and toUpdate -> Update).

The diffs also contain the `checkpoint_col` set to the start time of the DMS task. This is the time the DMS task started the full-load. Ongoing replication records will have `checkpoint_col` occurring after the DMS start time. This allows the archived records to be sorted by that column.


src/main/java/uk/gov/justice/digital/job/batchprocessing/ReloadDiffProcessor.java
```diff
@@ -0,0 +1,115 @@
...
+
+    public void createDiff(SourceReference sourceReference, String outputBasePath, Dataset<Row> raw, Dataset<Row> archive, Date reloadTime) {
+        String formattedStartTime = new SimpleDateFormat(DATE_TIME_PATTERN).format(reloadTime);
+        val windowFunction = Window.partitionBy(getKeyColumnsAsSeq(sourceReference)).orderBy(col(CHECKPOINT_COL).desc());
+
+        Dataset<Row> activeArchiveRecords = archive
+                .withColumn(RANK_COL, row_number().over(windowFunction))
+                .where(col(RANK_COL).eqNullSafe(lit(1)).and(col(OPERATION).notEqual(lit(Delete.getName()))))
+                .drop(RANK_COL)
+                .persist();
```
koladeadewuyi-moj, line 54: The diffs are computed against the active archive records. These are the most recent records for each primary key. Deletes are discarded because as they are no longer active.

tom-ogle-moj (approved):

PR #331, opened 2024-09-06
DPR2-1117: Data reconciliation job with current state counts.

tom-ogle-moj:
- Adds the new Data Reconciliation Job that runs for a specified input domain, outputs the results of the checks to logs and then fails the job is the checks have failed
- Implements the 1st check that Current State counts are the same across data stores (NOMIS, Structured, Curated)
- Checks counts are the same in the Operational Datastore if the ODS is enabled and if the table is managed by the ODS
- Checking DPS database counts instead of NOMIS for DPS input domains are out of scope of this story and will be covered in another
- Reworks some of the Operational Datastore code to be generic across JDBC databases so it can also be used for NOMIS
- DataReconciliationService is responsible for all data reconciliation
- CurrentStateCountService generates the current state counts


src/main/java/uk/gov/justice/digital/service/JDBCGlueConnectionDetailsService.java
```diff
@@ -0,0 +1,50 @@
...
+import uk.gov.justice.digital.datahub.model.JDBCCredentials;
+import uk.gov.justice.digital.datahub.model.JDBCGlueConnectionDetails;
+
+import java.util.Map;
+
+/**
+ * Responsible for retrieving JDBC details from AWS Glue Connections.
+ */
+@Singleton
+public class JDBCGlueConnectionDetailsService {
```
tom-ogle-moj, line 21: This is actually a class rename with a minor change as this was previously specific to the operational datastore.

src/main/java/uk/gov/justice/digital/service/NomisDataAccessService.java
```diff
@@ -0,0 +1,63 @@
...
+                connectionDetails.getUrl(),
+                connectionDetails.getJdbcDriverClassName(),
+                connectionDetails.getCredentials().getUsername(),
+                connectionDetails.getCredentials().getPassword()
+        );
+        logger.debug("Finished retrieving connection details for NOMIS");
+    }
+
+    @SuppressWarnings("java:S2077")
+    public long getTableRowCount(String tableName) {
```
tom-ogle-moj, line 48: It's much cheaper to do this than to read the data in to a Spark DataFrame and then count it. I've stuck with JDBC but if someone has strong opinions on using an ORM then I'm happy to switch it.

src/main/java/uk/gov/justice/digital/service/datareconciliation/model/CurrentStateCountTableResult.java
```diff
@@ -0,0 +1,46 @@
...
+/**
+ * Represents the results of running the total counts data reconciliation for a single "current state" table in DataHub
+ * replicated across multiple datastores
+ */
+@Getter
+public class CurrentStateCountTableResult {
+    private final long nomisCount;
+    private final long structuredCount;
+    private final long curatedCount;
+    // A null value indicates that there is no Operational DataStore count
```
tom-ogle-moj, line 14: I really wanted to use Optional to be more explicit but Java has a convention that Optional should not be used to represent absence vs presence in fields.

src/main/java/uk/gov/justice/digital/service/operationaldatastore/dataaccess/OperationalDataStoreConnectionDetailsService.java
```diff
@@ -1,56 +0,0 @@
...
-import uk.gov.justice.digital.datahub.model.OperationalDataStoreCredentials;
-
-import java.util.Map;
-
-/**
- * Responsible for retrieving details for connecting to the Operational DataStore.
- */
-@Singleton
-@Requires(property = "dpr.operational.data.store.write.enabled")
-public class OperationalDataStoreConnectionDetailsService {
```
tom-ogle-moj, line 23: This was actually renamed to JDBCGlueConnectonDetailsService and slightly modified

koladeadewuyi-moj (approved):

PR #332, opened 2024-09-11
DPR2-1160: try explicit system exit with error code when exception encountered during bootstrap

tom-ogle-moj:
- Fixes a bug where if a job fails while micronaut is creating objects whilst bootstrapping the dependency injection object graph then it would return a success exit code
- This resulted in jobs appearing as successful in glue when they should be reported as failed
- Introduces a class PicocliMicronautExecutor to run the job and handle exit code because Micronaut's PicocliRunner doesn't expose the execute method we need with ApplicationContext as a parameter
- We need ApplicationContext as a parameter because of the custom MicronautContext we are using to instantiate JobArguments, etc.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -49,6 +49,7 @@ ext {
     h2Version = '2.2.224'
     oracleJdbcVersion = '23.5.0.24.07'
     hikariCPVersion = '4.0.3'
+    junitSystemExitVersion = '1.1.2'
 }
 
 dependencies {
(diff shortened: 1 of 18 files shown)


tom-ogle-moj:
> Interesting find. LGTM. Would it be easy to add a test?

@gavriil-g-moj Well it is a bit tricky because of the `System.exit` call without adding in some heavy duty stuff like JMockit to test PicocliMicronautExecutor and still keep the code encapsulated. I have added a simple test though with a bit lighter weight approach to adding a test than using JMockit.

gavriil-g-moj:
> > Interesting find. LGTM. Would it be easy to add a test?
> 
> @gavriil-g-moj Well it is a bit tricky because of the `System.exit` call without adding in some heavy duty stuff like JMockit to test PicocliMicronautExecutor and still keep the code encapsulated. I have added a simple test though with a bit lighter weight approach to adding a test than using JMockit.

Yes totally understand it wouldn't be as straightforward to test this and it was only a suggestion if it did not require too much time.
Thanks for adding this test.

gavriil-g-moj (approved):
Thanks for adding a test.

koladeadewuyi-moj (approved):

PR #344, opened 2024-10-16
DPR2-1136: Report reconciliation to cloudwatch metrics.

tom-ogle-moj:
Initial spike.


tom-ogle-moj:
This approach would create a lot of expensive custom cloudwatch metric streams so I'm closing this in favour of my preferred approach.

PR #345, opened 2024-10-17
DPR2-1136: Report reconciliation failures to cloudwatch.

tom-ogle-moj:
- Reports reconciliaiton job failures to cloudwatch metrics
- Single dimension of domain (e.g. prisoner, activities) to keep custom metric costs low (only one custom metric per input domain)
- It reports a count of how many reconciliation checks failed

### DIFF
--- a/src/test/java/uk/gov/justice/digital/job/DataReconciliationJobTest.java
+++ b/src/test/java/uk/gov/justice/digital/job/DataReconciliationJobTest.java
@@ -60,11 +60,12 @@ class DataReconciliationJobTest {
 
     @Test
     @SuppressWarnings("java:S2699")
-    void runJobShouldNotExitWithErrorCodeWhenConfiguredNotToButResultIsFailure() {
+    void runJobShouldNotExitWhenConfiguredNotToButResultIsFailure() {
         when(dataReconciliationService.reconcileData(sparkSession)).thenReturn(results);
         when(results.isSuccess()).thenReturn(false);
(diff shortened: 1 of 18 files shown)


src/test/java/uk/gov/justice/digital/job/DataReconciliationJobTest.java
```diff
@@ -50,9 +50,20 @@ void runJobShouldReconcileData() {
...
+    }
+
+    @Test
+    @SuppressWarnings("java:S2699")
+    void runJobShouldNotExitWithErrorCodeWhenConfiguredNotToButResultIsFailure() {
+        when(dataReconciliationService.reconcileData(sparkSession)).thenReturn(results);
+        when(results.isSuccess()).thenReturn(false);
+        when(jobArguments.shouldReconciliationFailJobIfChecksFail()).thenReturn(false);
 
         underTest.runJob(sparkSession);
```
koladeadewuyi-moj, line 68: Is it worth asserting that the exit code is not > 0?

koladeadewuyi-moj (approved):

src/test/java/uk/gov/justice/digital/job/DataReconciliationJobTest.java
tom-ogle-moj: In this case the job should not exit at all so the test is checking that there is no System.exit. I will make the test name a bit clearer and add a comment

PR #356, opened 2025-01-16
DPR2-1633: Fix DMS Postgres Deletes going to violations.

tom-ogle-moj:
When the DMS processes deletes from Oracle it includes the body of the deleted row in the record. DMS with Postgres does not do this by default (only including the primary key columns and metadata columns added by DMS) and requires configuration on the database to do so.

This PR prevents deletes that do not include the body of the deleted row from failing validation and being sent to violations so that they can be processed normally.

To do this it:
- Only validates that required columns in the schema are included for operations which aren't a Delete
  - However, it still checks that primary key columns are included for Deletes as these are still required
- Adds an extra validation that the Operation column is not null


koladeadewuyi-moj (approved):

gavriil-g-moj (approved):
LGTM

PR #357, opened 2025-01-20
DPR2-1684: Delete old files and copy recent ones to archive

koladeadewuyi-moj:
Spark structured streaming commits file(s) to its checkpoint as soon as it picks them up. This leads to the scenario where check-pointed files may be moved by the archiving job before the processing is finished. This may lead to data loss i.e less records making it to the curated zone.

To avoid the aforementioned scenario, this PR modifies the `RawFileArchiveJob` so:
- It deletes raw files which have been check-pointed and are older than a configurable number of days (default is 2 days)
- It copies check-pointed files to the archive bucket without deleting the raw files


src/it/java/uk/gov/justice/digital/config/JobArgumentsIntegrationTest.java
```diff
@@ -349,6 +350,59 @@ public void shouldFailToGetFileTransferRetentionPeriodWhenGivenUnsupportedUnit(S
...
+            "seconds",
+            "HalfDays",
+            "Weeks",
+            "Months",
+            "Years",
+            "Decades",
+            "Centuries",
+            "Millennia",
+            "Eras",
+            "Forever"
```
tom-ogle-moj, line 396: This made me chuckle

tom-ogle-moj (approved):

PR #360, opened 2025-01-31
DPR2-1684: Ignore already archived files

koladeadewuyi-moj:
The archive job:
1. Archives raw files which have not been previously archived
2. Deletes all raw files which have been committed to the checkpoint and are older than 2 days (configurable)

To avoid wasting cost on re-archiving already archived files, the raw files are checked to ensure they do not have an existing key in the list of most recent archived files up until the last 2 days (configurable)

### DIFF
--- a/src/main/java/uk/gov/justice/digital/job/RawFileArchiveJob.java
+++ b/src/main/java/uk/gov/justice/digital/job/RawFileArchiveJob.java
@@ -7,12 +7,15 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import picocli.CommandLine;
 import uk.gov.justice.digital.config.JobArguments;
+import uk.gov.justice.digital.datahub.model.FileLastModifiedDate;
 import uk.gov.justice.digital.service.CheckpointReaderService;
 import uk.gov.justice.digital.service.ConfigService;
 import uk.gov.justice.digital.service.S3FileService;
 
 import javax.inject.Inject;
+import java.time.Clock;
 import java.time.Duration;
+import java.time.LocalDateTime;
 import java.util.ArrayList;
 import java.util.HashSet;
 import java.util.List;
@@ -31,6 +34,7 @@ public class RawFileArchiveJob implements Runnable {
     private final ConfigService configService;
     private final S3FileService s3FileService;
     private final CheckpointReaderService checkpointReaderService;
+    private final Clock clock;
     private final JobArguments jobArguments;
 
     @Inject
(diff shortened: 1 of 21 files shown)


src/main/java/uk/gov/justice/digital/job/RawFileArchiveJob.java
```diff
@@ -63,35 +64,44 @@ public void run() {
...
         Set<String> rawFiles = new HashSet<>(s3FileService
-                .listFilesForConfig(rawBucket, "", configuredTables, parquetFileRegex, Duration.ZERO));
+                .listFilesBeforePeriod(rawBucket, "", configuredTables, parquetFileRegex, Duration.ZERO));
+
+        Set<String> recentlyArchivedFiles = new HashSet<>(s3FileService
+                .listFilesAfterPeriod(archiveBucket, "", configuredTables, parquetFileRegex, archivedFilesCheckDuration));
 
-        List<String> filesToArchive = getCommittedFilesInProvidedFiles(committedFiles, rawFiles);
+        // Exclude the files which were already archived within the past specified period
+        List<String> filesToArchive = rawFiles.stream()
```
tom-ogle-moj, line 80: Probably a minor thing if we are not spending a lot of money on listing files in S3 but if I'm understanding this correctly then we list the bucket twice before taking the difference of all raw files and recently archived files. I wonder if instead of listing all files first and all files after the period seconds, then diffing them, could we instead just list all files before and within the period with a single list operation?

tom-ogle-moj (approved):
One comment but looks good

src/main/java/uk/gov/justice/digital/job/RawFileArchiveJob.java
koladeadewuyi-moj: Taking a look. Thanks

koladeadewuyi-moj: Thanks @tom-ogle-moj for your comment. I've pushed a commit which lists the raw files once and then uses the last modified date to determine which files are older than the retention period.

tom-ogle-moj: @koladeadewuyi-moj I think these two parts still call list twice?

```
Set<FileLastModifiedDate> rawFiles = new HashSet<>(s3FileService
                .listFilesBeforePeriod(rawBucket, "", configuredTables, parquetFileRegex, Duration.ZERO));
Set<String> recentlyArchivedFiles = new HashSet<>(s3FileService
                .listFilesAfterPeriod(archiveBucket, "", configuredTables, parquetFileRegex, archivedFilesCheckDuration)) ...
```

tom-ogle-moj (approved):

PR #365, opened 2025-05-04
DPR2-1774: Configure spark memory using worker type

koladeadewuyi-moj:
This PR creates a `dpr.adjust.spark.memory` flag to enable adjusting the memory used by the spark job to a higher value.

The flag defaults to false thereby maintaining default memory value i.e. the value injected by glue.

When the flag is set to true, the spark memory is adjusted to 0.75% instead of 0.625% of the Glue worker memory. This gives spark jobs which are spilling to disk more memory to work with.

### DIFF
--- a/src/main/java/uk/gov/justice/digital/job/CompactionJob.java
+++ b/src/main/java/uk/gov/justice/digital/job/CompactionJob.java
@@ -6,6 +6,7 @@ import org.apache.spark.sql.SparkSession;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import uk.gov.justice.digital.config.JobArguments;
+import uk.gov.justice.digital.config.JobProperties;
 import uk.gov.justice.digital.provider.SparkSessionProvider;
 import uk.gov.justice.digital.service.ConfigService;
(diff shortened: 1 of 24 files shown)


src/main/java/uk/gov/justice/digital/job/CompactionJob.java
```diff
@@ -51,25 +55,19 @@ public static void main(String[] args) {
...
     @Override
     public void run() {
-        try {
-            logger.info("Compaction running");
-            SparkSession spark = sparkSessionProvider.getConfiguredSparkSession(jobArguments);
-
-            String rootPath = jobArguments.getMaintenanceTablesRootPath();
-            int maxDepth = jobArguments.getMaintenanceListTableRecurseMaxDepth();
-            Optional<String> optionalConfigKey = jobArguments.getOptionalConfigKey();
+        SparkJobRunner.run("CompactionJob", jobArguments, properties, sparkSessionProvider, logger, this::runJob);
```
koladeadewuyi-moj, line 58: The spark job run logic was moved to the SparkJobRunner class to fix the SonarCloud code duplication warnings.

tom-ogle-moj (approved):

PR #423, opened 2026-07-15
DHS-662: Do not archive most recent checkpointed files

koladeadewuyi-moj:
To prevent data loss this PR does not archive the raw files associated with the most recent checkpoint thereby allowing the streaming job reprocess them when the it is restarted after a crash.

### DIFF
--- a/src/main/java/uk/gov/justice/digital/client/s3/S3CheckpointReaderClient.java
+++ b/src/main/java/uk/gov/justice/digital/client/s3/S3CheckpointReaderClient.java
@@ -64,7 +64,7 @@ public class S3CheckpointReaderClient {
 
     @NotNull
     private List<CheckpointFile> orderCheckpointFilesInReverseOrdering(List<FileLastModifiedDate> checkpointFiles) {
-        return checkpointFiles.stream()
+        List<CheckpointFile> processedCheckpointedFiles = checkpointFiles.stream()
                 .filter(checkpointFile -> !checkpointFile.key.toLowerCase().endsWith(".tmp"))
                 .map(checkpointFile -> {
                     Matcher matcher = checkpointFileRegexPattern.matcher(checkpointFile.key);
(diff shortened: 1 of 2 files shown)


src/main/java/uk/gov/justice/digital/client/s3/S3CheckpointReaderClient.java
```diff
@@ -76,6 +76,12 @@ private List<CheckpointFile> orderCheckpointFilesInReverseOrdering(List<FileLast
                     }
                 }).sorted(Collections.reverseOrder())
                 .collect(Collectors.toList());
+
+        // We ignore the most recent checkpoint file so the raw associated files with it do not get archived.
+        // This allows them to be reprocessed when the streaming job is restarted after a failure
+        if (!processedCheckpointedFiles.isEmpty()) processedCheckpointedFiles.remove(0);
```
tom-ogle-moj, line 82: It seems to me that this should work for avoiding the data loss. I suppose the consequence is that we only move files to raw archive as long as new data keeps coming? Perhaps that might mean that if a table is removed, or a domain is decommissioned, then we'd end up with some data left in raw that never makes it to the raw archive. I'm not sure how we could handle that?

koladeadewuyi-moj: We already copy all raw files which have not been previously archived to the archive bucket before getting to this stage. We do so for files whether they have been processed or not.
This stage is to prevent deletion of files which have not been processed from the raw zone.

koladeadewuyi-moj: The archive job needs to be updated so it considers all files instead of only the ones 2 days and older

tom-ogle-moj: `we already copy all raw files which have not been previously archived to the archive bucket before getting to this stage`

Ah right, so we copy all the files to the archive regardless of this checkpoint and then separately decide whether to delete from raw based on the checkpoints? So the files make it to the archive either way (whether it is in the latest checkpoint or not), it's just whether they are left in raw for reprocessing that this affects?

koladeadewuyi-moj: Yes. That's correct

